PR #8, opened 2022-10-12
Remove Ocelot dependencies in Cheetah

### LINKED ISSUE #4: Remove Ocelot dependencies
### 🚀 Feature Request 

Remove Ocelot as a dependency of Cheetah but keep functionalities like importing lattices and particle distributions.

### Motivation

Cheetah currently depends on Ocelot which causes long import times as well as issues with the installation (see #3 ).

cr-xu:
Fixes #4 

- [x]  Remove `astraBeam2particleArray` dependency
- [x]  Remove lattice conversion dependency

Particle distribution conversion validated with `astraBeam2particleArray`:

```python
import ocelot.adaptors as oca
import numpy as np
from cheetah.utils import from_astrabeam

parray = oca.astraBeam2particleArray("benchmark/astra/ACHIP_EA1_2021.1351.001")
particles, energy = from_astrabeam("benchmark/astra/ACHIP_EA1_2021.1351.001")

print("Maximal dev:", np.max((parray.rparticles.transpose() - particles)))
# > Maximal dev: 4.51461784622964e-16
```


cr-xu:
I thinke for now this would be sufficient. 

Installing Cheetah with Ocelot as optional dependency will happen once Ocelot is on PYPI.

PR #24, opened 2023-03-15
Add formatting check

cr-xu:
Add black and isort checking for CI/CD pipeline
Add pre-commit hooks and updated README.md


jank324 (approved):
We should probably also add flake8

Issue #2, opened 2022-10-07
Improved documentation


jank324:
### 📚 Documentation

The documentation needs some cleanup and extension. In particular, we should make sure that the API reference is clear and that there is a get-started guide.

### Motivation

The repo is now public and it should be easy for other people to use Cheetah.

jank324:
We also need to make the [readthedocs page](https://cheetah-accelerator.readthedocs.io/en/latest/) working along with improving the documentation.

jank324:
The page on readthedocs is now functional. What is missing is some more documentation to that page, including auto-generated documentation for all the functions and classes.

jank324:
Actually, the documentation is reasonably good now. So I'm closing this.

PR #29, opened 2023-03-17
Improved documentation

FelixTheilen:
This pull request is going to specifically adress issue #2 .
-The pull request introduces a new comment on the 'init' function. This will help finding the definitions for the parameters.
-Also a new comment on the 'transfer_map' function has been added, explaining the general purpose as well as the state vector and parameters, returns and raises.
-The parameter 'energy' has been removed from various classes. The parameter 'energy' is passed on through the beam passed to the segment. It is therefore not changeable by the user, when trying to control the beam, e.g. you can't call the function HorizontalCorrector and pass a parameter energy to this function. This parameter is defined, when the beam is fed in the Segment.
-'repr' has now a short comment describing the use of this function.
-Spelling mistakes have been corrected in accelerator.py and particles.py.

Unfortunately something brake down with the previous "Improved Documentation" pull request. This one should work now.

### DIFF
--- a/cheetah/accelerator.py
+++ b/cheetah/accelerator.py
@@ -50,6 +50,11 @@ class Element:
     is_skippable = True
 
     def __init__(self, name=None, device="auto"):
+        """
+        Used to define parameters. When defined they can be accessed through self.parameter via other function.
+        The function parameters change for each class. Since the parameters of this funvtion are always equivalent to the ones of the class
+        there will be no further description of the parameters.
+        """
         global ELEMENT_COUNT
         if name is not None:
             self.name = name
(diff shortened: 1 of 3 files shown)


cheetah/accelerator.py
```diff
@@ -46,6 +46,11 @@ class Element:
     is_skippable = True
 
     def __init__(self, name=None, device="auto"):
+        """
+        Used to define parameters. When defined they can be accessed through self.parameter via other function.
+        The function parameters change for each class. Since the parameters of this funvtion are always equivalent to the ones of the class
+        there will be no further description of the parameters.
+        """
```
jank324, line 53: Don't document `__init__`. Basically everything that function does should be obvious by the definition of the class already. Take a look at the [Stable Baselines3](https://github.com/DLR-RM/stable-baselines3) repository for a good example of how to do this well.

cheetah/accelerator.py
```diff
@@ -58,13 +63,41 @@ def __init__(self, name=None, device="auto"):
...
+        -------
+        NotImplementedError
+            If no transfer_map function is implemented for the given `Element` subclass.
+        """
         raise NotImplementedError
 
     def __call__(self, incoming):
         """
-        Track particles through the element.
+        Track particles through the element. Parameter and Particle Beams are treated differently, using an if statement.
```
jank324, line 98: The user does not need to know how the method works, i.e. that there is an `if` in it. Maybe just mention that it can handle `ParticleBeam` and `ParameterBeam`. It's also good practice to call them by there class names in reverse ticks.

cheetah/accelerator.py
```diff
@@ -123,11 +156,14 @@ def plot(self, ax, s):
...
         NotImplementedError
-            If not split function is implemented for the given `Element` subclass.
+            If no plot function is implemented for the given `Element` subclass.
         """
         raise NotImplementedError
 
     def __repr__(self):
+        """
+        This method is used to represent a class's objects as a string.
+        """
```
jank324, line 166: `__repr__` and all "magic" methods are a fixed component of Python, so what they do is well-defined and considered common knowledge / documented by python itself. There is therefore no need for a docstring here.

Issue #7, opened 2022-10-11
Establish testing


jank324:
### 🚀 Feature Request

Cheetah needs some sort of software tests that ensure everything works correctly and the computed results agree with other simulations such as Ocelot and ASTRA. Ideally, these would also test that Cheetah remains fast.

### Motivation

So far I have been using Notebooks in the `test` directory to roughly make sure that Cheetah still works when I changed things. This is obviously terrible, as it does not test everything, many of the Notebooks are out of date and it is a lot of work to run.

cr-xu:
Now that Felix is working on improving the code, establishing unit tests becomes more prominent so that we could have a CI/CD pipeline for PRs.

I'll try to add more tests in the upcoming weeks.

jank324:
Agreed! I've already talked to him that this is definitely something we should focus on next.

jank324:
We should use [pytest](https://docs.pytest.org/en/7.2.x/) for testing

jank324:
Btw, the `test` directory already has a couple of notebooks that I have used for testing. Many of them can probably be translated to "proper" pytest tests.

PR #33, opened 2023-03-29
Establish testing

FelixTheilen:
This pull request is adressing issue #7 .
There are still more tests needed to be implemented. This file is just the first. It is implementing the calculations done in the testless.ipynb file using pytest. This makes it a lot easier to run the test, as no manual comparison of the results is needed.


FelixTheilen:
Created a file which is implementing the actions performed in testtime.ipynb in pytest.

PR #34, opened 2023-03-31
Establish testing test_pycharm.py

FelixTheilen:
Creates a file which implements the testfile test.ipynb via pytest.

The following abilities are tested:
-import a `ParameterBeam` using .from_parameters
-import a `ParticleBeam` using .from_parameters
-using .transformed_to


jank324:
@FelixTheilen This pull request looks really good at first glance. The only thing I'm not sure about is the file name. The name should give some indication on what is tested by the tests in the file, so something like `test_beam_creation.py` (just an idea) would make more sense.

jank324:
Aside from that we just need to make sure all tests run (not just on @FelixTheilen's machine.

jank324:
@FelixTheilen Can you merge the most recent version of the `master` branch into this pull request? I've changed some things about the automatic testing that GitHub does, but it doesn't show up on your pull requests yet.

I think you can do this for this pull request first. If that gets the testing to show up properly, you can repeat it for the others as well.

PR #36, opened 2023-04-05
7 establish testing

cr-xu:
Addresses #7 

- Add pytest into CI workflow
- Remove obsolete test notebooks


cr-xu:
@FelixTheilen  Could you merge this branch afterwards into your development branches for #34 , #35 so that they also include pytest into  Github Actions

Issue #37, opened 2023-04-12
Conversion Error from Ocelot Monitor to Cheetah Screen


FelixTheilen:
The function `ocelot2cheetah` does not correctly convert Monitors written in Ocelot to Cheetah Screens. In the documentation it is said, that `ocelot.Monitor` elements containing the string "SCR" in their `id` are converted to Screen elements. Despite this the code is searching for the String "BSC".
Ideally the code should be changed in a way that older code containing the "BSC" identifier as well as newer code containing the "SCR" identifier can both be processed.
The documentaion should also be modified to explain this.

PR #38, opened 2023-04-13
Conversion error

FelixTheilen:
This pull request ia adressing the issue #37.
Now `Monitors` that are containing the string "BSC" as well as `Monitors` that are containing the string "SCR" are converted to Cheetah `Screens`. The Documentation has been updated accordingly.


cheetah/utils.py
```diff
@@ -109,7 +108,7 @@ def ocelot2cheetah(element, warnings=True):
         return acc.VerticalCorrector(element.l, element.angle, name=element.id)
     elif isinstance(element, oc.Cavity):
         return acc.Cavity(element.l, name=element.id)
-    elif isinstance(element, oc.Monitor) and "BSC" in element.id:
+    elif isinstance(element, oc.Monitor) and "BSC" in element.id or "SCR" in element.id:
```
jank324, line 111: A little thing here @FelixTheilen ... can you check precedence of `and` and `or`. I worry than as it is now, if `"SCR`" is in the name, it will completely ignore whether it is an instance of `oc.Monitor`.

FelixTheilen: It works in the desired way. `Drifts` containing `"SCR"` are not converted to `Screens`. However I added parentheses, as they will make the code more readable.

PR #41, opened 2023-05-03
Establish testing test_final.py

FelixTheilen:
This pull request addresses the issue establish testing #7 .

A segment of Ares is imported. A `ParameterBeam` and a `ParticleBeam` are sent through the segment. The outcome is compared to the supposed outcome.

Note: This test will always fail, as long as the pull request #38 Conversion Error is not implemented.


jank324:
@FelixTheilen there are some tests that fail on this pull request, do you know why?

Issue #42, opened 2023-05-16
Implement more lattice elements for cheetah


cr-xu:
Many lattice element types are still missing in Cheetah, and I'm working on adding these.

- [x] Bending magnets
  - [x] Sector Bend
  - [x] Rect Bend
- [x] Skew quadrupole
- [x] Solenoid 
- [x] Aperture

Need for higher-order tracking:

- [ ] Sextupole

jank324:
Pushing this up through mentions of @cr-xu @Hespe because I think we are getting to the point where a sextupole would be useful and then this issue could also be closed.

jank324:
One question would be if we should just wait for the Bmad-X sextupole to be added.

PR #46, opened 2023-05-17
Add more lattice elements in cheetah

cr-xu:
Added the following lattice elements:

- Dipole (Sector, Rectangular)
- Aperture (Rectangular, Elliptical)
  - blocks the beam with a large offset if tracking with `ParticleBeam`, otherwise ignored
  - needs to be activated by `self.is_active = True`
- Solenoid

Added `tilt` for `Quadrupole` , set tilt to pi/4 for skew-quadrupole

Moved some repeated transfer map calculations into a separate file: `track_method.py`

Addresses #42


jank324 (approved):
Nice!

Should we add at least some sanity tests for each of the new elements? I feel like it would be a good habit to get into, to create tests for every feature added and every bug fixed.

cr-xu:
Now added tests for Dipole (w & w/o fringe field), Solenoid, and Aperture

PR #56, opened 2023-07-12
9 loading and saving from lattices

cr-xu:
Implemented save and load via JSON format (according to the LatticeJSON convention: https://github.com/nobeam/latticejson)

Added a rough test of the functions.


cr-xu:
For a more comprehensive test, we need to check if the loaded lattice is identical to the original one. That would require implementing a `__eq__` method for all the lattice elements. @jank324  What's your opinion on this?

jank324:
Ah ... very nice! I like the idea. I wonder if there is any reason why we shouldn't just use the `@dataclass` decorator on (almost) all elements and get `__eq__` for free. 🤔

PR #57, opened 2023-07-13
Fix  `__eq__` check for Segment

cr-xu:
Changes:

- Overwrite the`__eq__` for Segment so that it checks that every element in the segment is identical.
- Updated the Ocelot lattice file for unique element id
- Updated the corresponding Cheetah file
- Updated test files accordingly.


cr-xu:
Sorry I forgot to mark #56 as still a work in progress 🙃

PR #63, opened 2023-09-03
Add twiss parameter calculations

jank324:
Adding twiss parameters of properties of the `Beam` class. This way they can easily be a computed from tracked `ParameterBeam` and `ParticleBeam`.

### DIFF
--- a/cheetah/particles.py
+++ b/cheetah/particles.py
@@ -328,27 +328,16 @@ class ParameterBeam(Beam):
         emittance_y: float = 0.0,
         energy: float = 1e8,
     ) -> "ParameterBeam":
-        sigma_x_squared = beta_x * emittance_x
-        cov_x_xp = -alpha_x * emittance_x
-        sigma_xp_squared = (emittance_x**2 + cov_x_xp) / sigma_x_squared
-        sigma_y_squared = beta_y * emittance_y
-        cov_y_yp = -alpha_y * emittance_y
-        sigma_yp_squared = (emittance_y**2 + cov_y_yp) / sigma_y_squared
+        sigma_x = np.sqrt(emittance_x * beta_x)
+        sigma_xp = np.sqrt(emittance_x * (1 + alpha_x**2) / beta_x)
+        sigma_y = np.sqrt(emittance_y * beta_y)
+        sigma_yp = np.sqrt(emittance_y * (1 + alpha_y**2) / beta_y)
 
-        return cls(
-            mu=torch.tensor([0, 0, 0, 0, 0, 0, 1], dtype=torch.float32),
-            cov=torch.tensor(
-                [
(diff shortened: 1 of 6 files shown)


jank324:
So this seems to work now. Here are a couple points we might need to think about:

- I think right now this doesn't work for `ParameterBeam`, but it absolutely should. The reason being that the emittance computation uses the actual particles:
```python
@property
def emittance_x(self) -> torch.Tensor:
    return torch.sqrt(
        self.sigma_x**2 * self.sigma_xp**2
        - torch.mean((self.xs - self.mu_x) * (self.xps - self.mu_xp)) ** 2
    )
```
the second part of this after the `-` this is inspired by Ocelot's 
```python
tws.xpx = np.mean((x - tws.x) * (px - tws.px))
```
which I thought is the same as `beam.sigma_x * beam.sigma_xp` in Cheetah, but apparently it's not.

- Is the twiss parameter computation from the Astra beam compared to Ocelot enough testing of this?
- We should probably also implement creating either type of beam from twiss parameters.

cr-xu:
> So this seems to work now. Here are a couple points we might need to think about:
> 
> * I think right now this doesn't work for `ParameterBeam`, but it absolutely should. The reason being that the emittance computation uses the actual particles:
> 
> ```python
> @property
> def emittance_x(self) -> torch.Tensor:
>     return torch.sqrt(
>         self.sigma_x**2 * self.sigma_xp**2
>         - torch.mean((self.xs - self.mu_x) * (self.xps - self.mu_xp)) ** 2
>     )
> ```

But this term $\sigma_{xx'}$ is actually just the `cor_x` in the covariance matrix used to initialize a `ParameterBeam`  
[c.f. here](https://github.com/desy-ml/cheetah/blob/377f4c0056557d4675adc2ecd5184edd4f794fe8/cheetah/particles.py#L224)
```python
cov=torch.tensor(
                [
                    [sigma_x**2, cor_x, 0, 0, 0, 0, 0],
                    [cor_x, sigma_xp**2, 0, 0, 0, 0, 0],
                    [0, 0, sigma_y**2, cor_y, 0, 0, 0],
                    [0, 0, cor_y, sigma_yp**2, 0, 0, 0],
                    [0, 0, 0, 0, sigma_s**2, cor_s, 0],
                    [0, 0, 0, 0, cor_s, sigma_p**2, 0],
                    [0, 0, 0, 0, 0, 0, 0],
                ],
                dtype=torch.float32,
            ),
```

cr-xu:
oops somehow the suggestion feature in GitHub messes up the indentation, now it's hopefully fixed

jank324:
Okay ... so I think all that's missing is to create beams from twiss parameters.

cr-xu:
Now the `from_twiss()` method should be fixed, the error was due to missing the cor_x and cor_y component, i.e. cov[0,1]=$\sigma_{xx'}$

Issue #61, opened 2023-09-03
Add MPS device


jank324:
Currently, Cheetah only considers `"cpu"` and `"cuda"` as possible devices. It would be nice to also make use of `"mps"` on Apple Silicon Macs. First, we need to test that all PyTorch functions used by Cheetah are already MPS-compatible. SB3, for example, doesn't used MPS yet because there are some incompatible functions.

jank324:
Just tried tracking a `ParameterBeam` through a `Quadrupole` using `torch=2.1.1`. I got

```
The operator 'aten::complex.out' is not currently implemented for the MPS device.
```

So PyTorch is not quite ready for Cheetah to support this yet.

jank324:
I just ran the `optimize_speed.ipynb` example notebook with MPS and I still get

```
NotImplementedError: The operator 'aten::complex.out' is not currently implemented for the MPS device. If you want this op to be added in priority during the prototype phase of this feature, please comment on https://github.com/pytorch/pytorch/issues/77764. As a temporary fix, you can set the environment variable `PYTORCH_ENABLE_MPS_FALLBACK=1` to use the CPU as a fallback for this op. WARNING: this will be slower than running natively on MPS.
```

So this operator is still not implemented.

jank324:
In the nightly build of PyTorch 2.4.0. With that, using an MPS device in Cheetah does work. However, it seems about 10x slower than running Cheetah on CPU with the stable build of PyTorch 2.2.1.

Testing was done using the vectorised branch (#116) and the `optimize_speed.ipynb` example notebook.

PR #69, opened 2023-09-08
Add device option for ParameterBeam

cr-xu:
- Now one can also specify the `device` for ParameterBeam
  - Added the according option in the doc string
  - Fixed the Element.__call__ device check

### DIFF
--- a/cheetah/accelerator.py
+++ b/cheetah/accelerator.py
@@ -97,17 +97,17 @@ class Element:
         if incoming is Beam.empty:
             return incoming
         elif isinstance(incoming, ParameterBeam):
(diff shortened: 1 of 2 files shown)


cr-xu:
We probably should include device error checks in tracking in the pytest. This requires then CUDA / MPS device (once #61 is addressed) in CI/CD runners. 
I don't know how to do it at the moment though...

jank324:
Ufff ... I wonder if especially MPS is even possible. I guess we can keep this in mind and otherwise postpone it until it comes up again. But testing would probably be good because clearly there were errors now that we didn't catch.

PR #75, opened 2023-09-14
Add CHANGELOG.md

cr-xu:
Add a `CHANGELOG.md` to keep track of the releases.

This should be merged before the 0.6.0 release.


jank324:
I think we can ignore the Python 3.0 test failure on this one. That problem should be fixed in #70, i.e. before this one will actually be merged. That means it should also go away when #70 is merged and this pull request is updated.

PR #78, opened 2023-09-15
Add newly implemented elements to Ocelot conversion

jank324:
We implemented some new Elements in Cheetah, but never added them to the Ocelot converter.


jank324:
@cr-xu can you in particular check `e1` and `e2` in the `RBend` conversion?

PR #80, opened 2023-09-15
Test Sphinx theme

jank324:
Switch docs to Awesome Sphinx Theme to test it.


cr-xu (approved):

cr-xu:
Somehow it shows merging is blocked for me
image
But this seems just a fast-forward. Also, we didn't change any branch protection rules, right?

jank324:
> Somehow it shows merging is blocked for me image But this seems just a fast-forward. Also, we didn't change any branch protection rules, right?

It says the same ... I really don't see why though. Maybe it's because the PR is very old?

I'm inclined to bypass the branch protections for this one if you also don't see any reasons why we shouldn't.

jank324:
We didn't change the branch protection rules recently. But given the age of the PR, it is possible we changed them since creating it.

cr-xu:
> It says the same for me ... I really don't see why though. Maybe it's because the PR is very old?
> 
> I'm inclined to bypass the branch protections for this one if you also don't see any reasons why we shouldn't.


Sounds good, let's just do that.

PR #86, opened 2023-09-26
Add charge in the Beam classes

cr-xu:
- Add total_charge to beam classes
  - `ParameterBeam` now has a `total_charge` attribute
  - `ParticleBeam` has now `q_array` which contains the charge of each particle in the `particles`. The `total_charge` is derived as `sum(q_array)`
  - Now the charge array will be also converted when importing from ASTRA or Ocelot beam.
- Update tests

The charge is tracked through the elements along with the beam. For now, the only element that affects the beam charge is the `Aperture` which removes the particle charges with large position offsets.


jank324:
I would propose giving a better name to `q_array` ... something more like `particle_charges`.

Also, can you add the feature to the changelog?

Otherwise, this looks good to me.

jank324:
Another thing that I'm just realising: Are the new charge properties properly tracked through all the elements?

jank324 (approved):

PR #88, opened 2023-09-28
Implement opt-in merging of elements

jank324:
I had an issue with the LCLS lattice, where the speed really wasn't great. This was likely due to that lattice having a huge number of elements. However, in my RL environment, I only ever tough a few of them. Most elements remain constant all the time.

I've implemented a new element type that just has a custom transfer map. This can be useful for all kinds of things actually. For my problem in particular, it is possible to take consecutive elements that won't be changed and combine their transfer maps into such an element.

In addition with some other minor optimisations (removing inactive Markers and so on) this gave me around a 20x speedup. I'm currently thinking about whether to also include those optimisations in this PR.

All these optimisations are (currently) opt-in, i.e. you create a new optimised segment by calling a method of the old segment.


jank324:
Also there is an issue right now that over the LCLS lattice I am not getting exactly the correct Twiss parameters. It's small enough to possibly be a numerical error, but large enough that it might also be a proper error. I am investigating.

jank324:
Remaining todos for this PR:

- [ ] Add convenience function for inactive marker removal
- [ ] Add  convenience function for removal of inactive zero-length elements
- [ ] Add convenience function for replacing inactive long elements with drifts
- [ ] Add tests
- [ ] Add optimisation example Notebook

cr-xu (changes requested):
I think `skippable_elements = []` should be moved outside of the if-clause / or add a reset also to after line 1746 in `tranfer_maps_merged`

Now if there's only one element between the `except_for` elements, the transfer_map keeps adding up everything:

```python
import torch
import cheetah

incoming_beam = cheetah.ParameterBeam.from_astra(
        "benchmark/astra/ACHIP_EA1_2021.1351.001"
    )
original_segment = cheetah.Segment(
        elements=[
            cheetah.Drift(length=torch.tensor(0.6)),
            cheetah.Quadrupole(length=torch.tensor(0.2), k1=torch.tensor(4.2), name="Q1"),
            cheetah.Drift(length=torch.tensor(0.4)),
            cheetah.HorizontalCorrector(
                length=torch.tensor(0.1), angle=torch.tensor(1e-4)
            , name="HCOR_1"),
            cheetah.Drift(length=torch.tensor(0.4)),
        ]
    )

merged_segment = original_segment.transfer_maps_merged(incoming_beam= incoming_beam, except_for=["Q1", "HCOR_1"])

print(merged_segment.elements)

# not equal
print(merged_segment.elements[2].transfer_map(energy=incoming_beam.energy))
print(original_segment.elements[2].transfer_map(energy=incoming_beam.energy))
```

cheetah/accelerator.py
```diff
@@ -1643,6 +1715,145 @@ def flattened(self) -> "Segment":
...
+                    merged_elements.append(skippable_elements[0])
+                    tracked_beam = skippable_elements[0].track(tracked_beam)
+                elif len(skippable_elements) > 1:  # i.e. we need to merge some elements
+                    merged_elements.append(
+                        CustomTransferMap.from_merging_elements(
+                            skippable_elements, incoming_beam=tracked_beam
+                        )
+                    )
+                    skippable_elements = []
+                    tracked_beam = merged_elements[-1].track(tracked_beam)
```
cr-xu, line 1754:
```suggestion
                    tracked_beam = merged_elements[-1].track(tracked_beam)
                skippable_elements = []
```

cr-xu (approved):

PR #90, opened 2023-10-04
Fix the transfer map of `Drift` and edge focusing of `Dipole`

cr-xu:
- Fix the R_56 component of the transfer map in `Drift`;
- Updated the test functions, now the difference should be less.
- Add R_56 component in `VerticalCorrector` and `HorizontalCorrector`
- Fix the edge focusing transfer maps for bends (e.g. `Dipole` `RBend` and `SBend`)
- Removed the large tolerances for result comparison in test functions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ None
 
 ### 🐛 Bug fixes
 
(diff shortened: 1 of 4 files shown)


cr-xu:
@jank324  This probably also helps to reduce the discrepancy between Bmad and cheetah tracking.

jank324 (approved):

PR #92, opened 2023-10-13
91 attributeerror screen object has no attribute reading

jank324:
The steps to solve this problem are:

- [ ] Write test to determine if that is actually the cause
- [ ] Copy solution from read_beam

Alternatively:

- [ ] Write tests to make sure that Cheetah is working fine and the issue is in the project using Cheetah


jank324:
So far it seems that Cheetah is behaving as expected and the problem is actually in the ARES Transverse Tuning repository.

jank324:
So it turns out the issue was that for `ParameterBeam` the code was wrong and crashed. Somehow this did not crash the kernel and the error was never reported. Instead the effect was that the next time `reading` was accessed, it was no longer available.

PR #96, opened 2023-10-19
95 nx tables import

jank324:
Implements reading from NX Tables, a CSV-like format used at ARES to save lattices and often considered as the ground truth there.

In this specific case this was needed to be able to read the lattice with the new geometry of the experimental area for Stage 4 v3.9, as there was no Ocelot conversion available for this.


jank324:
This is mostly done. The remaining todos are:

 - [ ] Check back with ARES guys if EA geometry change looks plausible
 - [ ] Remove development Notebook

jank324:
The second is loaded from NX Tables.

Screenshot 2023-10-20 at 08 51 34

cr-xu:
The code looks good to me.
I was actually hesitating on this PR because I don't know if this feature is too problem-specific. I.e. whether they should be put at the same level as other converters (widely used simulation codes), or in some specific folders (something like special_features with example notebooks explaining the usage cases).

jank324:
Hmmm ... so I did also wonder if it's so good to put ARES-specific code in here ... I actually don't know if these NX Tables are also used for other DESY accelerators.

For purely selfish reasons, I would like for this to be in Cheetah. It makes my code nicer.

While I don't think this gets in the way of other people using Cheetah, you are right that one day it might make Cheetah more bloated. This is not so much an issue with this converter being there than it is with this converter having to be maintained.

Right now, I don't think it makes Cheetah bloated. We could put it into an `experimental` (or similar) submodule of `converters` to highlight the fact that it is specific, so if it breaks in the future, it doesn't need an immediate fix. But it is also possible that in doing so, we bloat Cheetah now in an effort to avoid bloat in the future that might never come.

Difficult ... I will sleep over it.

jank324:
@cr-xu I think we should keep this as is.

For now the `converters` are not overflowing and we have no idea how much or little they will grow in the future. I think we shouldn't solve a problem now that we don't actually now yet we will have ... and then cross that bridge if and when we come to it.

Issue #113, opened 2023-12-11
Moving elements and beams to devices doesn't work


jank324:
It should be possible, like with any other PyTorch Module, to move Elements and Beams by calling

```python
element = element.to(device)
```

or

```python
element = element.cpu()
```

or

```python
element = element.cuda()
```

Unfortunately, this doesn't work right now.

jank324:
This will require registering all relevant variables as buffers:

https://stackoverflow.com/questions/60908827/why-pytorch-nn-module-cuda-not-moving-module-tensor-but-only-parameters-and-bu

cr-xu:
This is still not fixed, right?  So if I read the code correctly, now the device is specified only when one creates the `Element`, for changes afterwards one needs to shift every parameter to device manually...?

Are we still planning to integrate this feature, and how exactly... (explicitly registering the relevant parameters in buffer for all elements?)

jank324:
Right now it's bodged ... so you can do what you need to do in some way, but it's far from ideal.

This fix will be buffers (for everything that isn't a parameter). If we do it that way, everything should be movable like your average `nn.Linear` and so on.

However, this raises the question, what would happen if you do this ...

```python
quadrupole.k1 = torch.tensor([1.0])
```

... i.e. reassign a property that is now a buffer. Am I changing the values in the buffer? Or -- more likely -- am I completely overriding the buffer with a normal tensor and this assignment doesn't really work?

I think I saw you doing something in the BO example of cheetah-demos, where in GPyTorch that have some (hacky) work around for this. So we have to figure out how to make this work.

Either way, I think this will be an add-on on top of the batched execution #116 for a later version (0.7.1 or so).

PR #115, opened 2023-12-11
114 tests are failing on gpu machine

jank324:
## Description

This PR overhauls the way devices and types are handles by Elements and Beams. The idea is to make it more in line with how original PyTorch Modules like `nn.Linear` do it (https://pytorch.org/docs/stable/_modules/torch/nn/modules/linear.html#Linear). This also addresses the fact that according to (https://stackoverflow.com/questions/58926054/how-to-get-the-device-type-of-a-pytorch-module-conveniently) Modules shouldn't have overarching type and device properties because their parameters and buffers may be on different devices and of different types.

On a more practical note, this makes the entire device system much more robust and basically completes MPS support from Cheetah's side (#61 / not all operations we use are implemented in PyTorch for MPS yet).

A minor downside is that we no longer do automatic device selection. This does, however, make that part of Cheetah more predictable (I previously had a lot of code using Cheetah fix the device because weird things would happen).

This PR also prepares an eventual fix for #113.

## Motivation and Context

## Types of changes

## Checklist


cr-xu:
If I read the changes correctly, now we can safely switch between [float16, float32, float64] for elements, right?

jank324:
I haven't tested it (other than quickly trying to track through a quadrupole with MPS), but in theory, we should now be able to safely switch between all dtypes and devices, yes.

CHANGELOG.md
```diff
@@ -4,7 +4,7 @@
 
 ### 🚨 Breaking Changes
 
-None
+- The handling of `device` and `dtype` was overhauled. They might not behave as expected. `Element`s also no longer have a `device` attribute. (see #115) (@jank324)
```
cr-xu, line 8: I think it's worth mentioning explicitly in the documents that it actually changes the behaviour of element creation, i.e. one can now create elements conveniently as before `v.6.0`, without the requirement to wrap every parameter as tensor first

```python
dipole = cheetah.Dipole(length=0.1, angle=0.1, dtype=torch.float32)
```

cheetah/track_methods.py
```diff
@@ -56,27 +50,31 @@ def base_rmatrix(
...
-        device = "cuda" if torch.cuda.is_available() else "cpu"
+    device = length.device
+    dtype = length.dtype
 
-    tilt = tilt if tilt is not None else torch.tensor(0.0, device=device)
-    energy = energy if energy is not None else torch.tensor(0.0, device=device)
+    tilt = tilt if tilt is not None else torch.tensor(0.0, device=device, dtype=dtype)
+    energy = (
+        energy if energy is not None else torch.tensor(0.0, device=device, dtype=dtype)
+    )
```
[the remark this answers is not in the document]
jank324: I don't think we need to. This would add to the compute time (if only slightly) and the method is not intended to be used by the user. The Elements that use it should already have checked that the devices are correct, I think.

CHANGELOG.md
jank324: I think you are right, but I also think this should be tested before making that claim.

Issue #100, opened 2023-11-13
Batched execution


jank324:
It would be in the spirit of the Cheetah's applications, to enable batched execution. This would be, for example, that a quadrupole's strength is set to vector of strengths, which would then result in a 3D tensor of transfer maps and allow parallel evaluations at much higher speed.

cr-xu:
This is a tricky problem, which will be very helpful to have.

There are still two cases to be considered, e.g. batch evaluation of `incoming_beam` and batch evaluation of `segment_setting`

The `incoming_beam` situation is easier to consider and implement: we have a fixed `Segment` and `transfer_map`, we would just need to broadcast the transfer map multiplication properly with the batched/stacked incoming beam (which is already halfway done in the `ParticleBeam` case anyway).

The changing `segment_setting` case is more tricky, but also **more important**. Ultimately, we're modifying the network (transfer map) during the forward pass. One concern is whether we'll be able to keep the execution speed (somewhat) fast enough.

For bookkeeping purposes, the general feature I would like is the ability to perform fast calculation of outgoing beam parameters for one static incoming beam, and vectorized accelerator settings with the general form `[..., n_batch, n_dim]` where `n_dim` being the number of accelerator settings to be changed. 

Currently, I have to do something like: 
```python
def forward(self, X: torch.Tensor) -> torch.Tensor:
    input_shape = X.shape
    X = X.reshape(-1, n_dim)
    Y = torch.zeros(X.shape[:-1])
    for i, x in enumerate(X):
        self.segment.Q1.k1 = x[0] # Set the input parameters 
        ...
        # Track the beam
        out_beam = self.segment(self.incoming_beam)
        # Compute the objective
        obj = ...
        Y[i] = obj
    return Y.reshape(input_shape[:-1])
```

which more or less scale linearly (poorly):
```
i 1.0
CPU times: user 1.55 ms, sys: 1.06 ms, total: 2.61 ms
Wall time: 1.85 ms
i 10.0
CPU times: user 4.6 ms, sys: 14 µs, total: 4.61 ms
Wall time: 4.65 ms
i 100.0
CPU times: user 38.6 ms, sys: 266 µs, total: 38.9 ms
Wall time: 39.4 ms
i 1000.0
CPU times: user 335 ms, sys: 1.22 ms, total: 336 ms
Wall time: 347 ms
i 10000.0
CPU times: user 3.23 s, sys: 15.6 ms, total: 3.25 s
Wall time: 3.26 s
```

[...]

PR #116, opened 2024-01-04
Vectorised simulations

jank324:
**‼️ Do not merge before v0.6.2 has been released! This must go in a version after that.**

## Description

Makes all simulations in Cheetah vectorised, as you would expect in normal neural network model implementations.

## Motivation and Context

Vectorised simulations should allow for much faster parallel simulation of beams and settings. This is very similar to how you might run a neural network with batches, and helps both speed-based and gradient-based applications.

Closes #100.

## Types of changes

## Checklist


jank324:
Here is a tricky thing to consider: What effect does it have on gradients when we conditionally set some value in a computation to a constant. Is that effect bad? If so, how do we fix it?

E.g. `igamma2` in `Drift.transfer_map`.

jank324:
We also need to consider input dimension flexibility. For now, most of the reworked code probably assumes one input dimension for beams and one for settings. Maybe the final code can be more flexible, i.e. allow any number of input dimensions for both ... kind of like having n-dimensional batches.

jank324:
I've made the decision now, that the batch dimensions of the incoming beam and the settings must match. We should add a way to broadcast them, i.e. if a beam only has a single value or if a some elements only have single value, they are broadcast to the dimensions of the batched input. This should not be done if any of them has a different batch dimension, which should not be allowed and hence fail.

jank324:
Looking at the `optimize_speed.ipynb` example Notebook, it seems like this change currently slows down Cheetah between 2x and 4x.

jank324:
Some outstanding todos:
 - [ ] Make example Notebook on batched compuations
 - [ ] Add batching to speed optimisations example Notebook (cleaning it up)
 - [ ] Change gradient-based Notebook to use batches

jank324:
This will probably be a `v0.7` feature.

jank324:
The last commit solves #127.

@cr-xu we might need to look at this at some point. Small rounding errors sometimes result in the emittance computation taking `sqrt` of a negative number, or in an emittance of 0.0. I think both is not physically possible. For now this is solved with the clamping, but I wonder if there is another reason for this and the clamping only fixes the symptoms. Not at all urgent though.

cr-xu:
> The last commit solves #127.
> 
> @cr-xu we might need to look at this at some point. Small rounding errors sometimes result in the emittance computation taking `sqrt` of a negative number, or in an emittance of 0.0. I think both is not physically possible. For now this is solved with the clamping, but I wonder if there is another reason for this and the clamping only fixes the symptoms. Not at all urgent though.

At first sight, this is a price we pay for staying in the SI-Units. Emittances are just more commonly very small numbers, usually expressed in mm*mrad, etc. 

Since now we can choose the dtype freely, could it be resolved with double-precision...? Can you provide a MWE where the emittance goes to zero or negative numbers?

tests/test_drift.py
```diff
@@ -52,7 +52,7 @@ def test_device_like_torch_module():
     if not torch.cuda.is_available():
         return
 
-    element = cheetah.Drift(length=torch.tensor(0.2), device="cuda")
+    element = cheetah.Drift(length=torch.tensor([0.2]), device="cuda")
```
cr-xu, line 55: I understand it's difficult for broadcasting... but it seems really annoying that we have to do `length=torch.tensor([0.1])` for each element.

I would suggest that at least for default single lattice (non-batched) tracking we should allow users to just define a magnet without specifying the shape of the tensor..?  (e.g. provide some broadcasting from [] to [1]...?)

cr-xu: Or...maybe we mention it explicitly in the documentation that one can do. This is probably a easier way than putting `torch.tensor([...])` everywhere?
```python
segment = segment.broadcast(incoming_beam_batch_shape)
out_beam = segment.track(incoming_beam)
```

jank324: I agree. PyTorch actually does similar automatic broadcasting. But I would wait to integrate that until we know the computations and the general batched/vectorised design works.

jank324: So make sure it works first, then make sure it's nice to use. Right now I'm still chasing down at least one bug in the vectorised code.

jank324:
@cr-xu I think I messed up trying to merge your cavity fixes. Could you please have a look?

cr-xu:
> @cr-xu I think I messed up trying to merge your cavity fixes. Could you please have a look?

I believe it's properly fixed now.

jank324:
@cr-xu I have two things you might be able to answer:
 - I removed the voltage <= 0.0 case from the cavity code because I'm not sure if it makes sense to catch it in the vectorised example. Do you see any potential issue with this?
  - Do we have any tests that check for physical correctness of the cavity against known correct results or Ocelot or whatever else?

jank324:
Okay ... what else needs to be done?

Should we update the `cheetah-demos` repo before merging this ... kind of as a further test before putting this on master or should we just go for it?

cr-xu:
I just updated the BO with Cheetah prior example in the cheetah-demo in a separate branch (see [commit 9940072](https://github.com/desy-ml/cheetah-demos/commit/99400721c27ec7b683e697580979cba059410735))
The vectorized version works without any problem and shows a nice speed-up :)
I think the demos can be merged later back to the main branch, once this PR is merged.

The `broadcast` functionality is really handy to use. 👍

cheetah/accelerator.py
```diff
@@ -39,6 +39,8 @@ class Element(ABC, nn.Module):
     :param name: Unique identifier of the element.
     """
 
+    length: torch.Tensor = torch.zeros((1))
```
jank324, line 42: So, I don't really like the fact that length is a class property and how the lengthless classes assign `length` after being initialised in `broadcast`.

**BUT** I explored an alternative, where `length` is a parameter of `Element.__init__` or a read-only property of `Element`. However, this doesn't play nicely with the `length` property in `Segment` or the writable `length` properties in other elements that actually have a length.

So I think for now this is a good and simple solution. Most likely, these things will change again when we start integrating automatic broadcasting functionality for convenience.

jank324: Just thought I would leave this here for later reference. This is no call to action.

jank324:
@cr-xu I'm getting the _Merging is blocked_ message again. I would like to figure out this time what is causing it. The only restriction I could find in the settings is the one I set a long time ago that you cannot push directly to `master` without a PR. I think it has something to do with changes requested in a review not being resolved, so I requested a re-review from you to test this.

jank324:
Nevermind ... I fixed it. The branch locking option seems to have been the issue

jank324:
Okay ... I didn't really fix it, but I also don't get it.

The branch locking option seems to be what causes it to tell us that we cannot merge this PR. At first I didn't want to disable it because then I was able to push to master without a PR. It turns out though, that I can do that even when branch locking is turn on, presumably because I'm an admin. So I turned it off and now we can merge PRs again.

I don't like that I'm able to accidentally push to master and as far as I understand the current settings should prevent an admin to push to master ... but clearly they don't. I also googled this and it keeps pointing to an option "Include administrators", but I cannot seem to find that one. I feels a little bit like it may have been renamed or removed (?)

cr-xu (approved):
Looks good to me.
I'm not so sure about pushing `.vscode/ltex...` stuff though, but it's not a big issue.

Issue #137, opened 2024-03-27
Add space-charge effects in Cheetah


RemiLehe:
It would be valuable to add space-charge effects in Cheetah.

This could be done by adding space-charge to existing elements by using the standard technique of sub-stepping through an element (used in other accelerator codes). For instance, a drift without space-charge
```
Drift(length=L)
```
would be replaced (here with 3 substeps) by
```
Segment(
    elements=[
        Drift(length=L/6),
       SpaceChargeKick(length=L/3),
       Drift(length=L/3),
       SpaceChargeKick(length=L/3),
       Drift(length=L/3),
       SpaceChargeKick(length=L/3),
       Drift(length=L/6)
  ]
)
```

One could use the integrated Green function method for `SpaceChargeKick`, described here:
https://journals.aps.org/prab/abstract/10.1103/PhysRevSTAB.9.044204

This could then be benchmarked with examples from https://inspirehep.net/literature/1690374, as well as the example from the section "Free Expansion of a Cold Uniform Bunch" in https://inspirehep.net/files/178e98af93e0aaf3838a175835781aa0

ax3l:
Small comment: I'll take a look at  #142 :rocket:. Just for the above, technically, the space-charge kick is modeled like a thin element (no `lenghts=0`).

RemiLehe:
Yes, we are aware that the space charge kick does not add length to the element (see for instance [this comment](https://github.com/desy-ml/cheetah/pull/142/files#r1590642938)). 
In `SpaceChargeKick(length=L/3)`, `length` is the equivalent of [`slice_ds` in ImpactX](https://github.com/ECP-WarpX/impactx/blob/development/src/particles/spacecharge/GatherAndPush.cpp#L25).

As [suggested by @jank324](https://github.com/desy-ml/cheetah/pull/142/files#r1590642938), the above `length` argument should probably be renamed (e.g. `effec_length`), maybe as part of a [follow-up PR](https://github.com/desy-ml/cheetah/issues/160) to #142 .

ax3l:
Yes. In your link, it's not fully equivalent to our ds in ImpactX because we do not model space charge kicks conceptionally as _elements_. I stead, we slice elements (slice_ds=ds/N) and add collective effects between slices.

jank324:
Closing this because #142 was merged a while ago.

PR #142, opened 2024-04-03
Add space-charge to Cheetah

greglenerd:
## Description

Adds a new `SpaceChargeKick` element that enables adding space charge to any other elements.

## Motivation and Context

Implements #137.

See https://github.com/desy-ml/cheetah/issues/137

## Types of changes

## Checklist


cheetah/accelerator.py
```diff
@@ -299,6 +299,265 @@ def defining_features(self) -> list[str]:
...
+                    if k > 0:
+                        grid[i, j, 2 * ns - 2 - k] = G_value
+                    if i > 0 and j > 0:
+                        grid[2 * nx - 2 - i, 2 * ny - 2 - j, k] = G_value
+                    if j > 0 and k > 0:
+                        grid[i, 2 * ny - 2 - j, 2 * ns - 2 - k] = G_value
+                    if i > 0 and k > 0:
+                        grid[2 * nx - 2 - i, j, 2 * ns - 2 - k] = G_value
+                    if i > 0 and j > 0 and k > 0:
+                        grid[2 * nx - 2 - i, 2 * ny - 2 - j, 2 * ns - 2 - k] = G_value
```
RemiLehe, line 494: When filling the Green function by periodicity we want:
- The second element (in Python, index `1`) to be copied to the last element (in Python, index `2*nx-1`)
- The third element (in Python, index `2`) to be copied to the last-be-one element (in Python, index `2*nx-2`)
- etc.
Therefore, I think that the code needs to look like this:
```suggestion
                    if i > 0:
                        grid[2 * nx - i, j, k] = G_value
                    if j > 0:
                        grid[i, 2 * ny - j, k] = G_value
                    if k > 0:
                        grid[i, j, 2 * ns - k] = G_value
                    if i > 0 and j > 0:
                        grid[2 * nx - i, 2 * ny - j, k] = G_value
                    if j > 0 and k > 0:
                        grid[i, 2 * ny - j, 2 * ns - k] = G_value
                    if i > 0 and k > 0:
                        grid[2 * nx - i, j, 2 * ns - k] = G_value
                    if i > 0 and j > 0 and k > 0:
                        grid[2 * nx - i, 2 * ny - j, 2 * ns - k] = G_value
```

cheetah/accelerator.py
```diff
@@ -299,6 +307,356 @@ def defining_features(self) -> list[str]:
     def __repr__(self) -> str:
         return f"{self.__class__.__name__}(length={repr(self.length)})"
 
+class SpaceChargeKick(Element):
+    """
+    Simulates space charge effects on a beam.
```
greglenerd, line 312: these params might be changed: I was thinking of a way to integrate spacechargekicks as a setting (eg spacecharge = True) which would automatically build spacechargekick objects and incorporate them appropriately, as they don't represent any physical element of the accelerator. @cr-xu @jank324

cheetah/accelerator.py
```diff
@@ -299,6 +307,354 @@ def defining_features(self) -> list[str]:
...
+    :param length: Length of the element in meters.
```
greglenerd, line 313: As for now, time discretization is manually implemented through the lenght. If spacechargekick objects are built automatically, it might be interesting to play on this parameter.

cheetah/accelerator.py
```diff
@@ -299,6 +307,354 @@ def defining_features(self) -> list[str]:
...
+    # TODO: Implement splitting for SpaceCharge properly, for now just returns the
+    # element itself
+        return [self]
+
+
+    @property
+    def is_skippable(self) -> bool:
+        return False
+
+    def plot(self, ax: matplotlib.axes.Axes, s: float) -> None:
```
greglenerd, line 648: I need to understand better the plot functions of cheetah to implement something here. The only thing I know is there will be an issue given the fact that spacechargekick elements add 'artificial' lenght to the beam. @cr-xu @jank324

RemiLehe (changes requested):
Thanks a lot for this PR!

I added a few suggestions ; we can discuss them further offline.

cheetah/accelerator.py
jank324: @greglenerd So I think we need to be careful with the `length` property. In all existing elements, `length` represents the physical length of the beam pipe and Cheetah actually relies on this assumption in some places. I would suggest that `SpaceChargeKick.length = 0.0` all the time (because it doesn't add length to the beam pipe), and that the property of `SpaceChargeKick` currently named `length` is renamed to something like `effect_length`. Does that make sense?

Based on this, I think `SpaceChargeKick` could be plotted with something like `plt.axvspan` as a background highlight from `s - effect_length` to `s`.

cheetah/accelerator.py
jank324: I think at some point in the future, we should think about creating a `Effects` class and make `SpaceChargeKick` a subclass of that. But right now this doesn't make sense yet.

Nevertheless, here is how I think we should integrate adding space charge to elements right now in a way that can easily be extended in the future:

Basically what I would do is to add a method to the `Element` class that looks something like this:
```python
class Element:

...

    def with_space_charge(self, resolution: float = 0.01, *args, **kwargs) -> Segment:
        splits = self.split(resolution)
        splits_with_space_charge = # List of [split, sc, split, sc, ..., split, sc] ... probably itertools has a nice way to create this ... *args and **kwargs go into SpaceChargeKick
        return Segment(elements=splits_with_space_charge, name=f"{self.name}_with_space_charge")
```

This should end up looking similar to what @RemiLehe proposed in the very beginning. So if you do

```python
Drift(length=0.5, name="my_drift").with_space_chage(resolution=0.25, nx=64)
```

you get

```python
Segment(
    name="my_drift_with_space_charge",
    elements=[
        Drift(length=0.25), SpaceChargeKick(nx=64), Drift(length=0.25), SpaceChargeKick(nx=64)
    ],
)
```

This example is probably not quite correct in many ways, but I think it illustrates my idea.

The nice thing about doing it this way would be that it automatically works for all elements that implement `split` correctly, and that it would be relatively straightforward to extend this for other collective effects in the future.

Does this make sense?

cheetah/accelerator.py
```diff
@@ -299,6 +307,354 @@ def defining_features(self) -> list[str]:
     def __repr__(self) -> str:
         return f"{self.__class__.__name__}(length={repr(self.length)})"
 
+class SpaceChargeKick(Element):
+    """
+    Simulates space charge effects on a beam.
+    :param length: Length of the element in meters.
+    :param grid_precision: Number of grid points in each dimension.
+    :param grid_dimensions: Dimensions of the grid as multiples of sigma of the beam.
```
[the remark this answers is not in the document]
jank324: Maybe `num_grid_points_x`, `num_grid_points_y`, `num_grid_points_z` ... and `grid_extend_x`, `grid_extend_y`, `grid_extend_z`. I generally think names should be chose such that someone unfamiliar with the code doesn't always need to scroll back to the docstring to read and understand what variables are.

cheetah/accelerator.py
```diff
@@ -299,6 +307,354 @@ def defining_features(self) -> list[str]:
...
+
+    def __init__(
+        self,
+        length: Union[torch.Tensor, nn.Parameter],
+        nx: Union[torch.Tensor, nn.Parameter,int]=32,
+        ny: Union[torch.Tensor, nn.Parameter,int]=32,
+        ns: Union[torch.Tensor, nn.Parameter,int]=32,
+        dx: Union[torch.Tensor, nn.Parameter]=3,
+        dy: Union[torch.Tensor, nn.Parameter]=3,
+        ds: Union[torch.Tensor, nn.Parameter]=3,
```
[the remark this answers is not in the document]
jank324: Good question. I think their role is very similar to that of `Screen.resolution`. So I think no one will every want get the gradient with respect to any of them (?), meaning they don't have to be `nn.Parameter`. I think we could get away with making them `int` and `float`. But, in NumPy, if you interleave operations on NumPy types and Python types, computations slow down significantly. I don't know to what extend this is also an issue in PyTorch, but just to avoid it, I would make them all `torch.Tensor`.

cheetah/accelerator.py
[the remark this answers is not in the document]
jank324: Maybe it would also be easier to have a 3D tensor for each and then just call them `num_grid_points` and `grid_extend`?

I think both ways have their pros and cons. With `Quadrupole.misalignment` or `Screen.resolution` we currently use the one Tensor approach a lot in Cheetah. (One day it might be nice to also change of these to dataclasses, so you can do `num_grid_points.x` ... but that's for us to solve in the future, not now.)

cheetah/accelerator.py
```diff
@@ -299,6 +307,354 @@ def defining_features(self) -> list[str]:
...
+        self.nx = int(torch.as_tensor(nx, **factory_kwargs))
+        self.ny = int(torch.as_tensor(ny, **factory_kwargs))
+        self.ns = int(torch.as_tensor(ns, **factory_kwargs))
+        self.dx = torch.as_tensor(dx, **factory_kwargs) # in multiples of sigma
+        self.dy = torch.as_tensor(dy, **factory_kwargs)
+        self.ds = torch.as_tensor(ds, **factory_kwargs)
+
+
+    def _grid_shape(self) -> tuple[int,int,int]:
+        return (int(self.nx), int(self.ny), int(self.ns))
```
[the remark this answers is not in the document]
jank324: I actually like the name `grid_shape` in like with NumPy notation.

cheetah/accelerator.py
```diff
@@ -299,6 +307,354 @@ def defining_features(self) -> list[str]:
...
+        ns: Union[torch.Tensor, nn.Parameter,int]=32,
+        dx: Union[torch.Tensor, nn.Parameter]=3,
+        dy: Union[torch.Tensor, nn.Parameter]=3,
+        ds: Union[torch.Tensor, nn.Parameter]=3,
+        
+        name: Optional[str] = None,
+        device=None,
+        dtype=torch.float32,
+    ) -> None:
+        factory_kwargs = {"device": device, "dtype": dtype}
```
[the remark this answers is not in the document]
jank324: It looks to me like you are using it correctly. It's just a helper variable to make sure all tensors are on matching devices and have matching dtypes.

cheetah/accelerator.py
```diff
@@ -299,6 +307,354 @@ def defining_features(self) -> list[str]:
...
 
+class SpaceChargeKick(Element):
+    """
+    Simulates space charge effects on a beam.
+    :param length: Length of the element in meters.
+    :param grid_precision: Number of grid points in each dimension.
+    :param grid_dimensions: Dimensions of the grid as multiples of sigma of the beam.
+    :param name: Unique identifier of the element.
+    """
+
```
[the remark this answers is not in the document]
jank324: I think this should be in the doctring of the `SpaceChargeKick` class.

cheetah/accelerator.py
```diff
@@ -299,6 +307,354 @@ def defining_features(self) -> list[str]:
...
+                incoming.energy,
+                particle_charges=incoming.particle_charges,
+                device=particles.device,
+                dtype=particles.dtype,
+            )
+        else:
+            raise TypeError(f"Parameter incoming is of invalid type {type(incoming)}")
+
+
+    def split(self, resolution: torch.Tensor) -> list[Element]:
```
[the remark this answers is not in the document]
jank324: I think the correct behaviour for an abstract element like `SapceChargeKick` is that `split` actually returns `[self]`.

cheetah/accelerator.py
```diff
@@ -299,6 +307,354 @@ def defining_features(self) -> list[str]:
...
+            return incoming
+        elif isinstance(incoming, ParticleBeam):
+            particles = self._cheetah_to_moments(incoming)
+            forces = self._read_forces(incoming)
+            dt = self._delta_t(incoming)
+            particles[:,1] += forces[:,0]*dt
+            particles[:,3] += forces[:,1]*dt
+            particles[:,5] += forces[:,2]*dt
+            particles = self._moments_to_cheetah(particles, incoming)
+            return ParticleBeam(
```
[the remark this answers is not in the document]
jank324: We always modify a copy, i.e. if you do

```python
outgoing = segment.track(incoming)
```

the user can expect that `incoming` is not changed.

jank324:
Weirdly all GitHub Actions except for the docs build are currently not showing up on this PR. We need to keep an eye on this.

cheetah/accelerator.py
```diff
@@ -45,7 +53,7 @@ def __init__(self, name: Optional[str] = None) -> None:
         self.name = name if name is not None else generate_unique_name()
 
     def transfer_map(self, energy: torch.Tensor) -> torch.Tensor:
-        r"""
+        """
```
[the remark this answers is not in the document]
greglenerd: I think it was a change from a previous commit with an error. Because we don't need the r.

cheetah/accelerator.py
```diff
@@ -299,6 +307,382 @@ def defining_features(self) -> list[str]:
...
+        # Shape: (n_particles, 8, 3)
+        weights = 1 - torch.abs(normalized_pos.unsqueeze(1) - surrounding_indices)  
+        # Shape: (n_particles, 8, 3)
+        cell_weights = weights.prod(dim=2)  # Shape: (n_particles, 8)
+=======
+        offsets = torch.tensor([[0, 0, 0], [0, 0, 1], [0, 1, 0], [0, 1, 1], [1, 0, 0], [1, 0, 1], [1, 1, 0], [1, 1, 1]])
+        surrounding_indices = cell_indices.unsqueeze(-2) + offsets  # Shape: (..., n_particles, 8, 3)
+        weights = 1 - torch.abs(normalized_pos.unsqueeze(-2) - surrounding_indices)  # Shape: (..., n_particles, 8, 3)
+        cell_weights = weights.prod(dim=-1)  # Shape: (..., n_particles, 8) - product of shapes along x, y and z
+>>>>>>> ea8bac2b8ec1ed6dbb9c88078775c8a243aafa2b
```
RemiLehe, line 416: @greglenerd Could you fix the above `<<<<<` , `>>>>>`

cheetah/accelerator.py
```diff
@@ -299,6 +307,382 @@ def defining_features(self) -> list[str]:
...
+        moments[:,5] = (gamma-gammaref*torch.ones(N))/(betaref*gammaref)
+        return moments
+
+    def _compute_forces(self, beam: ParticleBeam, cell_size, grid_dimensions) -> torch.Tensor:
+        """
+<<<<<<< HEAD
+        Compute the forces on the particles due to the space charge.
+=======
+        Interpolates the space charge force from the grid onto the macroparticles
+>>>>>>> ea8bac2b8ec1ed6dbb9c88078775c8a243aafa2b
```
RemiLehe, line 592: Could you fix the `>>>>`

cheetah/accelerator.py
```diff
@@ -299,6 +307,382 @@ def defining_features(self) -> list[str]:
...
+        num_grid_points_y: Union[torch.Tensor, nn.Parameter,int]=32,
+        num_grid_points_s: Union[torch.Tensor, nn.Parameter,int]=32,
+        grid_extend_x: Union[torch.Tensor, nn.Parameter]=3,
+        grid_extend_y: Union[torch.Tensor, nn.Parameter]=3,
+        grid_extend_s: Union[torch.Tensor, nn.Parameter]=3,
+        name: Optional[str] = None,
+        device=None,
+        dtype=torch.float32,
+    ) -> None:
+        factory_kwargs = {"device": device, "dtype": dtype}
```
RemiLehe, line 354:
```suggestion
        self.factory_kwargs = {"device": device, "dtype": dtype}
```

cheetah/accelerator.py
```diff
@@ -299,6 +307,382 @@ def defining_features(self) -> list[str]:
...
+        Track particles through the element. The input must be a `ParticleBeam`.
+        :param incoming: Beam of particles entering the element.
+        :return: Beam of particles exiting the element.
+        """
+        if incoming is Beam.empty or incoming.particles.shape[0] == 0:
+            return incoming
+        elif isinstance(incoming, ParticleBeam):
+            grid_dimensions = self._compute_grid_dimensions(incoming)
+            cell_size = 2*grid_dimensions / torch.tensor(self.grid_shape)
+            dt = self.length / (c*self._betaref(incoming))
```
RemiLehe, line 648:
```suggestion
            dt = self.length / (c*self._betaref(incoming))
            # Copy the array of coordinates to avoid modifying the incoming beam
            outcoming_particles = torch.empty_like(incoming.particles)
            outcoming_particles[...] = incoming.particles
            outcoming = ParticleBeam(
                outcoming_particles,
                incoming.energy,
                particle_charges=incoming.particle_charges,
                device=particles.device,
                dtype=particles.dtype,
            )
```
And then pass `outcoming` to all the function that were taking `incoming` + `return outcoming`

tests/test_space_charge_kick.py
```diff
@@ -0,0 +1,98 @@
...
+        cheetah.SpaceChargeKick(L/3),
+        cheetah.Drift(L/3),
+        cheetah.SpaceChargeKick(L/3),
+        cheetah.Drift(L/6)
+    ]
+    )
+    outgoing_beam = segment_space_charge.track(incoming)
+
+    # Final beam properties
+    sig_xo = outgoing_beam.sigma_x
```
jank324, line 57: So I don't really like giving these shorter aliases like `sig_xo`. It mostly just obscures what the code is doing. I would suggest renaming `outgoing_beam` to `outgoing` and then using `outgoing.sigma_x` for better readability.

tests/test_space_charge_kick.py
```diff
@@ -0,0 +1,98 @@
...
+    Tests that the incoming beam is not modified when calling the track method.
+    """
+
+    incoming_beam = cheetah.ParticleBeam.from_parameters(
+        num_particles=torch.tensor([10000]),
+        sigma_xp=torch.tensor([2e-7]),
+        sigma_yp=torch.tensor([2e-7]),
+    )
+    # Initial beam properties
+    incoming_particles0 = incoming_beam.particles
```
jank324, line 77: Why not name this like `incoming_beam_before`. This way no comments would be needed because the variable name would be self-explanatory.

Also, you should compare the entire beam because there is more to a beam than its particles.

cheetah/particles.py
```diff
@@ -1018,6 +1018,140 @@ def from_twiss(
             device=device,
             dtype=dtype,
         )
+    
+    @classmethod
+    def uniform_3d_ellispoid(
```
jank324, line 1023: This function has since been added on `master` and actually merged onto this branch as well. So this copy here should be removed and only the one from `master` should remain.

cheetah/accelerator.py
```diff
@@ -20,17 +20,25 @@
 from cheetah.track_methods import base_rmatrix, misalignment_matrix, rotation_matrix
 from cheetah.utils import UniqueNameGenerator
 
+#Constants
```
ax3l, line 23: PEP-8, probably
```suggestion
# Constants
```

cheetah/accelerator.py
```diff
@@ -20,17 +20,25 @@
...
+c = torch.tensor(constants.speed_of_light)
+J_to_eV = torch.tensor(physical_constants["electron volt-joule relationship"][0])
 generate_unique_name = UniqueNameGenerator(prefix="unnamed_element")
-
+elementary_charge= torch.tensor(constants.elementary_charge)
```
ax3l, line 27: PEP-8, probably
```suggestion
elementary_charge = torch.tensor(constants.elementary_charge)
```

tests/test_space_charge_kick.py
```diff
@@ -0,0 +1,98 @@
...
+    )
+    outgoing_beam = segment_space_charge.track(incoming)
+
+    # Final beam properties
+    sig_xo = outgoing_beam.sigma_x
+    sig_yo = outgoing_beam.sigma_y
+    sig_so = outgoing_beam.sigma_s
+
+    torch.set_printoptions(precision=16)
+    assert torch.isclose(sig_xo,2*sig_xi,rtol=2e-2,atol=0.0)
```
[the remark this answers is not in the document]
ax3l: Since `sig_xo` should not be close to zero, `atol` should not be [needed](https://pytorch.org/docs/stable/generated/torch.isclose.html). Maybe `rtol` just needs to be slightly larger to account for small changes in the order of summed elements from charge deposition (not sure about the level of parallelism on this in Cheetah).

tests/test_space_charge_kick.py
```diff
@@ -0,0 +1,98 @@
...
+    # Compute section lenght
+    kappa= 1+(torch.sqrt(torch.tensor(2))/4)*torch.log(3+2*torch.sqrt(torch.tensor(2)))
+    Nb = total_charge/cheetah.elementary_charge
+    L=beta*gamma*kappa*torch.sqrt(R0**3/(Nb*cheetah.electron_radius))
+
+    segment_space_charge = cheetah.Segment(
+    elements=[
+        cheetah.Drift(L/6),
+        cheetah.SpaceChargeKick(L/3),
+        cheetah.Drift(L/3),
```
ax3l, line 47: Looking at this now, this is a pretty uncommon/confusing way to express instantaneous kicks in a beamline.

The simplest way might be to make this a "thin" (zero-sized) element with `L=0` enforced.

cheetah/accelerator.py
```diff
@@ -327,6 +335,431 @@ def defining_features(self) -> list[str]:
     def __repr__(self) -> str:
         return f"{self.__class__.__name__}(length={repr(self.length)})"
 
+class SpaceChargeKick(Element):
+    """
+    Applies the effect of space charge over a length `length`, on the **momentum**
+    (i.e. divergence and energy spread) of the beam.
```
ax3l, line 341: This first part uncommon/confusing from my experience. I would make this a thin element (L=0) and apply it instantaneously multiple times; slice the element to apply it in.

The rest below is great (nice doc string!! :sparkling_heart: )

cheetah/accelerator.py
```diff
@@ -299,6 +307,354 @@ def defining_features(self) -> list[str]:
...
+        return 2*grid_dimensions / torch.tensor(grid_shape)
+    
+    def _gammaref(self,beam: ParticleBeam) -> torch.Tensor:
+        return beam.energy / rest_energy
+    
+    def _betaref(self,beam: ParticleBeam) -> torch.Tensor:
+        gamma = self._gammaref(beam)
+        if gamma == 0:
+            return torch.tensor(1.0)
+        return torch.sqrt(1 - 1 / gamma**2)
```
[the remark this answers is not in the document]
ax3l: Let's fix it or add an inline TODO comment on this? I already see someone comparing a 30 MeV proton beam and ending up very confused :)

cheetah/accelerator.py
ax3l: An alternative thought: One could also create sub- or mixing classes for thin and thick (L>0) elements. We currently use mixin classes in ImpactX:
https://github.com/ECP-WarpX/impactx/tree/development/src/particles/elements/mixin
e.g. [Drift](https://github.com/ECP-WarpX/impactx/blob/24.04/src/particles/elements/Drift.H#L27-L32) vs. thin [Multipole](https://github.com/ECP-WarpX/impactx/blob/24.04/src/particles/elements/Multipole.H#L27-L31)

Space charge kicks could be thin elements :)

tests/test_space_charge_kick.py
cr-xu: Also I think users also shouldn't type in this themselves when constructing segments. 
Later there should be some method in any component `Segment` to add space charge effect (or any other effect modeled as a instantaneous kick) either making use of the split method  or specifying the `ds` explicitly, 
See also above (https://github.com/desy-ml/cheetah/pull/142#discussion_r1590657538)

jank324: See also #160.

cheetah/accelerator.py
RemiLehe: Yes, I agree that calling this `length` is confusing.

Note that, in the current PR, the slicing is done by hand by the user. For instance, in the CI test,
```
Segment(
    elements=[
        Drift(length=L/6),
       SpaceChargeKick(length=L/3),
       Drift(length=L/3),
       SpaceChargeKick(length=L/3),
       Drift(length=L/3),
       SpaceChargeKick(length=L/3),
       Drift(length=L/6)
  ]
)
```
corresponds to breaking the element `Drift(length=L)` into 3 slices.

But as discussed in the PR and in issues, this should not be the final interface and their should be a follow-up PR to introduce a more user-friendly interface.

ax3l: I agree, for this PR that is ok and we can finalize this part in a follow-up as discussed.

cheetah/accelerator/space_charge_kick.py
```diff
@@ -0,0 +1,595 @@
...
+            )
+
+        # End of loop over batch
+        inv_cell_volume = 1 / (cell_size[:, 0] * cell_size[:, 1] * cell_size[:, 2])
+
+        return (
+            charge * inv_cell_volume[:, None, None, None]
+        )  # Normalize by the cell volume
+
+    def _integrated_potential(self, x, y, s) -> torch.Tensor:
```
jank324, line 187: Can we add type annotations?

cheetah/accelerator/space_charge_kick.py
```diff
@@ -0,0 +1,595 @@
...
+        """
+        Computes the electrostatic potential using the Integrated Green Function method
+        as in http://dx.doi.org/10.1103/PhysRevSTAB.9.044204.
+        """
+
+        r = torch.sqrt(x**2 + y**2 + s**2)
+        G = (
```
jank324, line 194: I guess it's called `G` in the original method. Capital single letter variable names are not very pythonic and can sometimes be confusing. Though it might still be clearer if the name from the original paper is used. So it could alternatively be called `integrated_potential`? Or we just circumvent this by returning without naming it.

cheetah/accelerator/space_charge_kick.py
```diff
@@ -0,0 +1,595 @@
...
+        # Copy the original charge_density values to the beginning of the new tensor
+        new_charge_density[
+            :,
+            : charge_density.shape[1],
+            : charge_density.shape[2],
+            : charge_density.shape[3],
+        ] = charge_density
+        return new_charge_density
+
+    def _IGF(self, beam: ParticleBeam, cell_size) -> torch.Tensor:
```
jank324, line 229:
```suggestion
    def _integrated_green_function(self, beam: ParticleBeam, cell_size) -> torch.Tensor:
```

jank324: Also, can we add a type annotation to `cell_size`?

cheetah/accelerator/space_charge_kick.py
```diff
@@ -0,0 +1,595 @@
...
+            + self._integrated_potential(
+                x_grid - 0.5 * dx, y_grid - 0.5 * dy, s_grid + 0.5 * ds
+            )
+            - self._integrated_potential(
+                x_grid - 0.5 * dx, y_grid - 0.5 * dy, s_grid - 0.5 * ds
+            )
+        )
+
+        # Initialize the grid with double dimensions
+        green_func = torch.zeros(
```
jank324, line 286: `green_func` is probably also a bit confusing here because it's not a function, but whatever that function return, right?

jank324 (changes requested):

cheetah/__init__.py
```diff
@@ -13,6 +13,7 @@
     Screen,
     Segment,
     Solenoid,
+    SpaceChargeKick,
```
RemiLehe, line 16:
```suggestion
    SpaceChargeKick, # add a comment
```

cheetah/accelerator/space_charge_kick.py
```diff
@@ -0,0 +1,595 @@
...
+            - 0.5 * y**2 * torch.atan(x * s / (y * r))
+            - 0.5 * x**2 * torch.atan(y * s / (x * r))
+            + y * s * torch.asinh(x / torch.sqrt(y**2 + s**2))
+            + x * s * torch.asinh(y / torch.sqrt(x**2 + s**2))
+            + x * y * torch.asinh(s / torch.sqrt(x**2 + y**2))
+        )
+        return G
+
+    def _array_rho(
+        self, beam: ParticleBeam, cell_size, grid_dimensions
```
cr-xu, line 205:
```suggestion
        self, beam: ParticleBeam, cell_size: torch.Tensor, grid_dimensions: torch.Tensor
```

cheetah/accelerator/space_charge_kick.py
```diff
@@ -0,0 +1,595 @@
...
+        # Return the physical potential
+        return potential[
+            :,
+            : charge_density.shape[1] // 2,
+            : charge_density.shape[2] // 2,
+            : charge_density.shape[3] // 2,
+        ]
+
+    def _E_plus_vB_field(
+        self, beam: ParticleBeam, cell_size, grid_dimensions
```
cr-xu, line 362:
```suggestion
        self, beam: ParticleBeam, cell_size: torch.Tensor, grid_dimensions: torch.Tensor
```

cheetah/accelerator/space_charge_kick.py
```diff
@@ -0,0 +1,595 @@
...
+        gamma = torch.sqrt(1 + (p / (electron_mass * c)) ** 2)
+        moments[:, :, 1] = moments[:, :, 1] / p0[:, None]
+        moments[:, :, 3] = moments[:, :, 3] / p0[:, None]
+        moments[:, :, 4] = -moments[:, :, 4] / betaref[:, None]
+        moments[:, :, 5] = (gamma - gammaref * torch.ones(gamma.shape)) / (
+            (betaref * gammaref)[:, None]
+        )
+
+    def _compute_forces(
+        self, beam: ParticleBeam, cell_size, grid_dimensions
```
cr-xu, line 437:
```suggestion
        self, beam: ParticleBeam, cell_size: torch.Tensor, grid_dimensions: torch.Tensor
```

cheetah/accelerator/space_charge_kick.py
```diff
@@ -0,0 +1,595 @@
...
+        return beam.energy / rest_energy
+
+    def _betaref(self, beam: ParticleBeam) -> torch.Tensor:
+        gamma = self._gammaref(beam)
+        if gamma == 0:
+            return torch.tensor(1.0)
+        return torch.sqrt(1 - 1 / gamma**2)
+
+    def _deposit_charge_on_grid(
+        self, beam: ParticleBeam, cell_size, grid_dimensions
```
cr-xu, line 114:
```suggestion
        self, beam: ParticleBeam, cell_size: torch.Tensor, grid_dimensions: torch.Tensor
```

cheetah/particles/particle_beam.py
```diff
@@ -327,7 +461,7 @@ def uniform_3d_ellipsoid(
             ]
             if argument is not None
         ]
-        shape = not_nones[0].shape if len(not_nones) > 0 else torch.Size([1])
+        shape = not_nones[0].size() if len(not_nones) > 0 else torch.Size([1])
```
cr-xu, line 464:
```suggestion
        shape = not_nones[0].shape if len(not_nones) > 0 else torch.Size([1])
```

cheetah/particles/particle_beam.py
```diff
@@ -269,6 +269,140 @@ def from_twiss(
...
+            zip(radius_x.view(-1), radius_y.view(-1), radius_s.view(-1))
+        ):
+            particles[i] = generate_uniform_3d_ellispoid_particles(
+                num_particles, r_x, r_y, r_s
+            )[0]
+        parray.particles = particles.view(*shape, num_particles, 7)
+        parray.particles.to(device=device, dtype=dtype)
+
+        return parray
+
```
cr-xu, line 405:
```suggestion

```

cheetah/accelerator/space_charge_kick.py
```diff
@@ -0,0 +1,595 @@
...
+        # Copy the original charge_density values to the beginning of the new tensor
+        new_charge_density[
+            :,
+            : charge_density.shape[1],
+            : charge_density.shape[2],
+            : charge_density.shape[3],
+        ] = charge_density
+        return new_charge_density
+
+    def _integrated_green_function(self, beam: ParticleBeam, cell_size: torch.Tensor) -> torch.Tensor:
```
cr-xu, line 229:
```suggestion
    def _integrated_green_function(
        self, beam: ParticleBeam, cell_size: torch.Tensor
    ) -> torch.Tensor:
```

cheetah/accelerator/space_charge_kick.py
cr-xu:
```suggestion
        self, beam: ParticleBeam, cell_size: torch.Tensor, grid_dimensions: torch.Tensor
```

cheetah/accelerator/space_charge_kick.py
```diff
@@ -0,0 +1,597 @@
...
+
+    def _solve_poisson_equation(
+        self, beam: ParticleBeam, cell_size, grid_dimensions
+    ) -> torch.Tensor:  # works only for ParticleBeam at this stage
+        """
+        Solves the Poisson equation for the given charge density, using FFT convolution.
+        """
+        charge_density = self._array_rho(beam, cell_size, grid_dimensions)
+        charge_density_ft = torch.fft.fftn(charge_density, dim=[1, 2, 3])
+        integrated_green_function = self._IGF(beam, cell_size)
```
cr-xu, line 346:
```suggestion
        integrated_green_function = self._integrated_green_function(beam, cell_size)
```

tests/test_space_charge_kick.py
```diff
@@ -0,0 +1,112 @@
...
+    rest_energy = torch.tensor(
+        constants.electron_mass
+        * constants.speed_of_light**2
+        / constants.elementary_charge
+    )
+    elementary_charge = torch.tensor(constants.elementary_charge)
+    electron_radius = torch.tensor(physical_constants["classical electron radius"][0])
+    gamma = energy / rest_energy
+    beta = torch.sqrt(1 - 1 / gamma**2)
+    incoming = cheetah.ParticleBeam.uniform_3d_ellispoid(
```
cr-xu, line 32:
```suggestion
    incoming = cheetah.ParticleBeam.uniform_3d_ellipsoid(
```

cr-xu:
There appears to be also [random failing of CI](https://github.com/desy-ml/cheetah/actions/runs/9370530781/job/25797719359?pr=142) due to numerical issues?

RemiLehe:
@cr-xu Yes, the tests where failing randomly, due to stochastic fluctuations in the initial particle distribution. In order to avoid this, I fixed the random seed in the test.

cheetah/accelerator/space_charge_kick.py
```diff
@@ -0,0 +1,595 @@
...
+        grid_extend_y: Union[torch.Tensor, nn.Parameter] = 3,
+        grid_extend_s: Union[torch.Tensor, nn.Parameter] = 3,
+        name: Optional[str] = None,
+        device=None,
+        dtype=torch.float32,
+    ) -> None:
+        self.factory_kwargs = {"device": device, "dtype": dtype}
+        super().__init__(name=name)
+        self.length_effect = torch.as_tensor(length_effect, **self.factory_kwargs)
+        self.length = torch.as_tensor(length, **self.factory_kwargs)
```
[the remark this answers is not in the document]
RemiLehe: Yes, maybe as a follow-up PR that also addresses https://github.com/desy-ml/cheetah/issues/160

cheetah/accelerator/space_charge_kick.py
RemiLehe: Sounds good.

cheetah/accelerator/space_charge_kick.py
```diff
@@ -0,0 +1,597 @@
...
+        )  # Normalize by the cell volume
+
+    def _integrated_potential(self, x, y, s) -> torch.Tensor:
+        """
+        Computes the electrostatic potential using the Integrated Green Function method
+        as in http://dx.doi.org/10.1103/PhysRevSTAB.9.044204.
+        """
+
+        r = torch.sqrt(x**2 + y**2 + s**2)
+        G = (
```
RemiLehe, line 194:
```suggestion
        integrated_potential = (
```

cheetah/accelerator/space_charge_kick.py
```diff
@@ -0,0 +1,597 @@
...
+            -0.5 * s**2 * torch.atan(x * y / (s * r))
+            - 0.5 * y**2 * torch.atan(x * s / (y * r))
+            - 0.5 * x**2 * torch.atan(y * s / (x * r))
+            + y * s * torch.asinh(x / torch.sqrt(y**2 + s**2))
+            + x * s * torch.asinh(y / torch.sqrt(x**2 + s**2))
+            + x * y * torch.asinh(s / torch.sqrt(x**2 + y**2))
+        )
+        return G
```
RemiLehe, line 202:
```suggestion
        return integrated_potential
```

cheetah/accelerator/space_charge_kick.py
RemiLehe: Agreed. How about `green_func_values`?

cheetah/accelerator/space_charge_kick.py
```diff
@@ -0,0 +1,607 @@
...
+    ) -> torch.Tensor:
+        """
+        Deposits the charge density of the beam onto a grid, using the nearest
+        grid point method and weighting by the distance to the grid points.
+        Returns a grid of charge density in C/m^3.
+        """
+        charge = torch.zeros((self.n_batch,) + self.grid_shape, **self.factory_kwargs)
+
+        # Loop over batch dimension
+        for i_batch in range(self.n_batch):
```
RemiLehe, line 133:
```suggestion
        # Loop over samples in one batch (does vectorization)
        for i_batch in range(self.batch_size):
```

cheetah/accelerator/space_charge_kick.py
```diff
@@ -0,0 +1,607 @@
...
+                outgoing_particles,
+                incoming.energy,
+                particle_charges=incoming.particle_charges,
+                device=incoming.particles.device,
+                dtype=incoming.particles.dtype,
+            )
+            # Flatten the batch dimensions
+            # (to simplify later calculation, is undone at the end of `track`)
+            outgoing.particles.reshape((-1, outgoing.num_particles, 7))
+            self.n_batch = outgoing.particles.shape[0]
```
RemiLehe, line 571:
```suggestion
            self.batch_size = outgoing.particles.shape[0]
```

cheetah/accelerator/space_charge_kick.py
```diff
@@ -0,0 +1,607 @@
...
+
+    def _deposit_charge_on_grid(
+        self, beam: ParticleBeam, cell_size: torch.Tensor, grid_dimensions: torch.Tensor
+    ) -> torch.Tensor:
+        """
+        Deposits the charge density of the beam onto a grid, using the nearest
+        grid point method and weighting by the distance to the grid points.
+        Returns a grid of charge density in C/m^3.
+        """
+        charge = torch.zeros((self.n_batch,) + self.grid_shape, **self.factory_kwargs)
```
RemiLehe, line 130:
```suggestion
        charge = torch.zeros((self.batch_size,) + self.grid_shape, **self.factory_kwargs)
```

cheetah/accelerator/space_charge_kick.py
```diff
@@ -0,0 +1,607 @@
...
+        Allocates a 2x larger array in all dimensions (to perform Hockney's method),
+        and copies the charge density in one of the "quadrants".
+        """
+        grid_shape = self.grid_shape
+        charge_density = self._deposit_charge_on_grid(beam, cell_size, grid_dimensions)
+        new_dims = tuple(dim * 2 for dim in grid_shape)
+
+        # Create a new tensor with the doubled dimensions, filled with zeros
+        new_charge_density = torch.zeros(
+            (self.n_batch,) + new_dims, **self.factory_kwargs
```
RemiLehe, line 228:
```suggestion
            (self.batch_size,) + new_dims, **self.factory_kwargs
```

cheetah/accelerator/space_charge_kick.py
```diff
@@ -0,0 +1,607 @@
...
+        num_grid_points_x, num_grid_points_y, num_grid_points_s = self.grid_shape
+
+        # Create coordinate grids
+        x = torch.arange(num_grid_points_x, **self.factory_kwargs)
+        y = torch.arange(num_grid_points_y, **self.factory_kwargs)
+        s = torch.arange(num_grid_points_s, **self.factory_kwargs)
+        ix_grid, iy_grid, is_grid = torch.meshgrid(x, y, s, indexing="ij")
+        x_grid = (
+            ix_grid[None, :, :, :] * dx[:, None, None, None]
+        )  # Shape: [n_batch, nx, ny, nz]
```
RemiLehe, line 262:
```suggestion
        )  # Shape: [batch_size, nx, ny, nz]
```

cheetah/accelerator/space_charge_kick.py
```diff
@@ -0,0 +1,610 @@
...
+        super().__init__(name=name)
+        self.length_effect = torch.as_tensor(length_effect, **self.factory_kwargs)
+        self.length = torch.as_tensor(length, **self.factory_kwargs)
+        self.grid_shape = (
+            int(num_grid_points_x),
+            int(num_grid_points_y),
+            int(num_grid_points_s),
+        )
+        self.grid_extend_x = torch.as_tensor(grid_extend_x, **self.factory_kwargs)
+        # in multiples of sigma
```
jank324, line 87: I don't really understand this comment

cheetah/accelerator/space_charge_kick.py
```diff
@@ -0,0 +1,610 @@
...
+        )
+
+        # Scale the gradients with lorentz factor
+        grad_x = -igamma2[:, None, None, None] * grad_x
+        grad_y = -igamma2[:, None, None, None] * grad_y
+        grad_s = -igamma2[:, None, None, None] * grad_s
+
+        return grad_x, grad_y, grad_s
+
+    def _cheetah_to_moments(self, beam: ParticleBeam) -> torch.Tensor:
```
jank324, line 412: This function currently doesn't return anything ... it seems to do things inplace 🤔

cheetah/accelerator/space_charge_kick.py
```diff
@@ -0,0 +1,610 @@
...
+        beta = torch.sqrt(1 - 1 / gamma**2)
+        p = gamma * electron_mass * beta * c
+        moments[:, :, 1] = p0[:, None] * moments[:, :, 1]
+        moments[:, :, 3] = p0[:, None] * moments[:, :, 3]
+        moments[:, :, 4] = -betaref[:, None] * moments[:, :, 4]
+        moments[:, :, 5] = torch.sqrt(
+            p**2 - moments[:, :, 1] ** 2 - moments[:, :, 3] ** 2
+        )
+
+    def _moments_to_cheetah(self, beam: ParticleBeam) -> torch.Tensor:
```
jank324, line 433: This function currently doesn't return anything ... it seems to do things inplace 🤔

cheetah/accelerator/space_charge_kick.py
```diff
@@ -0,0 +1,610 @@
...
+
+        return grad_x, grad_y, grad_s
+
+    def _cheetah_to_moments(self, beam: ParticleBeam) -> torch.Tensor:
+        """
+        Converts the Cheetah particle beam parameters to the moments in SI units used
+        in the space charge solver.
+        """
+        moments = beam.particles
+        gammaref = self._gammaref(beam)
```
jank324, line 424: Should `gammaref` just be a property of the beam? (And have a better name?)

cheetah/accelerator/space_charge_kick.py
jank324: Actually, `moments` could also be a property of `ParticleBeam` I guess?

cheetah/accelerator/space_charge_kick.py
```diff
@@ -0,0 +1,610 @@
...
+            outgoing = ParticleBeam(
+                outgoing_particles,
+                incoming.energy,
+                particle_charges=incoming.particle_charges,
+                device=incoming.particles.device,
+                dtype=incoming.particles.dtype,
+            )
+            # Flatten the batch dimensions
+            # (to simplify later calculation, is undone at the end of `track`)
+            outgoing.particles.reshape((-1, outgoing.num_particles, 7))
```
jank324, line 583: This line currently doesn't do anything because the output is not "caught"

jank324:
Okay ... the gradient compute in-place error is fixed, but in ([Replace obvious in-place operations with out-of-place alternatives](https://github.com/desy-ml/cheetah/pull/142/commits/fd0be7a99ea173adc1e19c4339f20e6e48c8e5f9) I appear to have introduced a bug that just copies some lines of the `particles` tensor rather than returning the one updated with the space charge effects.

cheetah/accelerator/space_charge_kick.py
```diff
@@ -0,0 +1,655 @@
...
+        for i_batch in range(self.batch_size):
+            # Get particle positions and charges
+            particle_positions = moments[i_batch, :, [0, 2, 4]]
+            particle_charges = beam.particle_charges[i_batch]
+            normalized_positions = (
+                particle_positions[:, :] + grid_dimensions[i_batch, None, :]
+            ) / cell_size[i_batch, None, :]
+
+            # Find indices of the lower corners of the cells containing the particles
+            cell_indices = torch.floor(normalized_positions).type(torch.long)
```
jank324, line 151: Is long a good idea?

jank324:
I've fixed the vectorisation now as well. However, there are two remaining issues, which I will highlight in comments below.

tests/test_space_charge_kick.py
```diff
@@ -0,0 +1,182 @@
...
+            cheetah.Drift(section_length / 3),
+            cheetah.SpaceChargeKick(section_length / 3),
+            cheetah.Drift(section_length / 3),
+            cheetah.SpaceChargeKick(section_length / 3),
+            cheetah.Drift(section_length / 6),
+        ]
+    )
+    outgoing = segment.track(incoming)
+
+    assert torch.isclose(outgoing.sigma_x, 2 * incoming.sigma_x, rtol=2e-2, atol=0.0)
```
jank324, line 66: This comparison fails since my last commit. I seem to have introduced a small error into the computation. Maybe someone else can check if you can find it.

cheetah/accelerator/space_charge_kick.py
```diff
@@ -0,0 +1,684 @@
...
+        valid_mask = (
+            (idx_x >= 0)
+            & (idx_x < grid_shape[0])
+            & (idx_y >= 0)
+            & (idx_y < grid_shape[1])
+            & (idx_s >= 0)
+            & (idx_s < grid_shape[2])
+        )
+
+        # TODO: Is there a better alternative to this loop? (maybe torch.vmap?)
```
jank324, line 563: This loop is a bit of a hack to achieve functionality like vectorisation, but of course it's not true vectorisation.

More importantly, it breaks the gradient computations. Ideally, we would find a way to properly vectorise it like the rest. I couldn't immediately come up with something, but that might also due to my lack of understanding exactly what is going on. In the worst case, we could try to fix it with `torch.vmap`.

jank324:
I think these two are the last remaining things before @RemiLehe can investigate the gradient-computations. Before merging, there are a few refactoring items we should probably take care of, but I would like to discuss them first.

cheetah/accelerator/space_charge_kick.py
```diff
@@ -84,146 +86,162 @@ def __init__(
...
+            & (idx_s >= 0)
+            & (idx_s < self.grid_shape[2])
+        )
+
+        # Accumulate the charge contributions
+        repeated_charges = beam.particle_charges.repeat_interleave(
+            repeats=8, dim=-1
+        )  # Shape:(..., 8 * num_particles)
+        values = (cell_weights.flatten(start_dim=-2) * repeated_charges)[valid_mask]
+        charge[..., idx_x[valid_mask], idx_y[valid_mask], idx_s[valid_mask]] += values
```
cr-xu, line 187: I tracked down the issue to this line, which is not producing the same result as the previous inplace `index_put_`

jank324: I surprised right now that that's still in there ... I thought I reverted that 🤔

jank324: @cr-xu do you happen to have caught how the values differed?

cr-xu: So a simple `+=` doesn't handle values added to repeated indices, this line basically only adds charges once to a grid, causing the SC force to be way smaller thant it should be

jank324: Just putting the original implementation here for reference:

```python
charge[i_batch].index_put_(
    (idx_x[valid_mask], idx_y[valid_mask], idx_s[valid_mask]),
    values,
    accumulate=True,
)
```

cr-xu: I think [`scatter_add_`](https://pytorch.org/docs/stable/generated/torch.Tensor.scatter_add_.html#torch.Tensor.scatter_add_) would be a replacement, which has the backpass capability

jank324: I'm trying to understand the differences between [`scatter_add_`](https://pytorch.org/docs/stable/generated/torch.Tensor.scatter_add_.html#torch.Tensor.scatter_add_) and [`index_put_`](https://pytorch.org/docs/stable/generated/torch.Tensor.scatter_add_.html#torch.Tensor.index_put_). The problem with the original `index_put_` is that it doesn't yet work on the vectorised tensors and I'm not quite sure how it would be made to do that.

jank324: I think the crux is that the vector indices need to be correctly included in the index. If the it done, we can probably just revert to the old `index_put_` line.

jank324:
Notes from our meeting about remaining things to do:
 - [ ] Implemented flatten hack to fix vectorisation and its gradients
 - [ ] Make `betaref` and `gammaref` properties of `ParticleBeam`
 - [ ] Make `moments a property of `ParticleBeam` and rename it (@RemiLehe you wanted to check what Ocelot calls it?)

cheetah/accelerator/space_charge_kick.py
```diff
@@ -0,0 +1,684 @@
...
+        )
+
+        # Get particle positions
+        particle_positions = moments[..., [0, 2, 4]]
+        normalized_positions = (
+            particle_positions + grid_dimensions.unsqueeze(-2)
+        ) / cell_size.unsqueeze(-2)
+
+        # Find indices of the lower corners of the cells containing the particles
+        cell_indices = torch.floor(normalized_positions).type(torch.long)
```
jank324, line 143: Why `long` and not `int`?

cheetah/particles/particle_beam.py
```diff
@@ -722,6 +724,72 @@ def transformed_to(
...
+        beam.particles[..., 1] = moments[..., 1] / p0.unsqueeze(-1)
+        beam.particles[..., 3] = moments[..., 3] / p0.unsqueeze(-1)
+        beam.particles[..., 4] = -moments[..., 4] / beam.relativistic_beta.unsqueeze(-1)
+        beam.particles[..., 5] = (gamma - beam.relativistic_gamma.unsqueeze(-1)) / (
+            (beam.relativistic_beta * beam.relativistic_gamma).unsqueeze(-1)
+        )
+
+        return beam
+
+    def to_moments(self) -> torch.Tensor:
```
jank324, line 765: We wanted to think of a better name for this.

cheetah/particles/particle_beam.py
```diff
@@ -722,6 +724,72 @@ def transformed_to(
             dtype=dtype,
         )
 
+    @classmethod
+    def from_moments(
```
jank324, line 728: We wanted to think of a better name for this.

jank324:
So most issues are fixed. The only one I'm not really getting on top of is that the computation is still slightly wrong.

cr-xu:
> So most issues are fixed. The only one I'm not really getting on top of is that the computation is still slightly wrong.

I managed to fix that, but it seems that I can't push to the upstream branch 
```
error: failed to push some refs to 'github.com:greglenerd/cheetah.git'
```

jank324:
> > So most issues are fixed. The only one I'm not really getting on top of is that the computation is still slightly wrong.
> 
> I managed to fix that, but it seems that I can't push to the upstream branch
> 
> ```
> error: failed to push some refs to 'github.com:greglenerd/cheetah.git'
> ```

Huh ... odd ... I thought because you are Maintainer, you should have permissions for that. As a quick fix, maybe you can push to a temporary branch and then I merge it into the upstream one manually?

cr-xu:
> > > So most issues are fixed. The only one I'm not really getting on top of is that the computation is still slightly wrong.
> > 
> > 
> > I managed to fix that, but it seems that I can't push to the upstream branch
> > ```
> > error: failed to push some refs to 'github.com:greglenerd/cheetah.git'
> > ```
> 
> Huh ... odd ... I thought because you are Maintainer, you should have permissions for that. As a quick fix, maybe you can push to a temporary branch and then I merge it into the upstream one manually?

Okay I pushed the fix to [fix-compute-forces-in-space-charge](https://github.com/desy-ml/cheetah/tree/fix-compute-forces-in-space-charge)

jank324:
> > > > So most issues are fixed. The only one I'm not really getting on top of is that the computation is still slightly wrong.
> > > 
> > > 
> > > I managed to fix that, but it seems that I can't push to the upstream branch
> > > ```
> > > error: failed to push some refs to 'github.com:greglenerd/cheetah.git'
> > > ```
> > 
> > 
> > Huh ... odd ... I thought because you are Maintainer, you should have permissions for that. As a quick fix, maybe you can push to a temporary branch and then I merge it into the upstream one manually?
> 
> Okay I pushed the fix to [fix-compute-forces-in-space-charge](https://github.com/desy-ml/cheetah/tree/fix-compute-forces-in-space-charge)

Is merged!

jank324:
I've gone through now and resolved all comments that are already dealt with or out-of-date. Here is what remains to do before we can merge:

 - [ ] Make sure the plotting is correct (https://github.com/desy-ml/cheetah/pull/142#issuecomment-2178491914)
 - [ ] Consider if there is a better name for `moments` (https://github.com/desy-ml/cheetah/pull/142#discussion_r1645149168)
 - [ ] Answer the question about the use of `torch.long` (https://github.com/desy-ml/cheetah/pull/142#discussion_r1644755047)
 - [ ] Implement `defining_features` (https://github.com/desy-ml/cheetah/pull/142#discussion_r1626159492)
 - [ ] Implement `__repr__` (https://github.com/desy-ml/cheetah/pull/142#discussion_r1626160494)
 
 If you think of anything else, please edit this post and add it here.

RemiLehe:
@cr-xu @jank324 This is awesome! Thanks a lot for fixing the space-charge test, along with differentiation and vectorization! You guys are so fast!

Regarding a replacement for `moments`, here are a few suggestions:
- `cheetah_coords_to_xp`
- `cheetah_coords_to_x_y_z_px_py_pz`
- `canonical_coords_to_xyz_pxpypz`
Any preferences? (For reference, [here](https://github.com/ocelot-collab/ocelot/blob/master/ocelot/cpbd/coord_transform.py#L57) is the corresponding Ocelot function, but may want to use a different name.)

Regarding plotting: could we simply plot a vertical line (as if the space-charge kick was a thin element), instead of a rectangle?

Regarding `torch.long`, I think we could remove it and use regular `int`s instead. I will push a corresponding commit.

jank324:
> Regarding a replacement for `moments`, here are a few suggestions:
> 
> * `cheetah_coords_to_xp`
> * `cheetah_coords_to_x_y_z_px_py_pz`
> * `canonical_coords_to_xyz_pxpypz`
>   Any preferences? (For reference, [here](https://github.com/ocelot-collab/ocelot/blob/master/ocelot/cpbd/coord_transform.py#L57) is the corresponding Ocelot function, but may want to use a different name.)

I have no opinion regarding the `xp`, `x_y_z_px_py_pz` or `xyz_pxpypz` question. But I would have the function just `to_xyz_pxpypz` (or any of the three options) because in the end it would be used as

```python
moments = cheetah_beam.to_xyz_pxpypz()
another_cheetah_beam = ParameterBeam.from_xyz_pxpypz(moments, ...)
```

> Regarding plotting: could we simply plot a vertical line (as if the space-charge kick was a thin element), instead of a rectangle?

A vertical line might actually be more intuitive, yes. Still, can someone explain why we do

```python
segment = cheetah.Segment(
    elements=[
        cheetah.Quadrupole(quad_length / 6),
        cheetah.SpaceChargeKick(quad_length / 3),
        cheetah.Quadrupole(quad_length / 3),
        cheetah.SpaceChargeKick(quad_length / 3),
        cheetah.Quadrupole(quad_length / 3),
        cheetah.SpaceChargeKick(quad_length / 3),
        cheetah.Quadrupole(quad_length / 6),
    ]
)
```

rather than

```python
segment = cheetah.Segment(
    elements=[
        cheetah.Quadrupole(quad_length / 3),
        cheetah.SpaceChargeKick(quad_length / 3),
        cheetah.Quadrupole(quad_length / 3),
        cheetah.SpaceChargeKick(quad_length / 3),
        cheetah.Quadrupole(quad_length / 3),
        cheetah.SpaceChargeKick(quad_length / 3),
    ]
)
```

?

cr-xu:
> > Regarding a replacement for `moments`, here are a few suggestions:
> > 
> > * `cheetah_coords_to_xp`
> > * `cheetah_coords_to_x_y_z_px_py_pz`
> > * `canonical_coords_to_xyz_pxpypz`
> >   Any preferences? (For reference, [here](https://github.com/ocelot-collab/ocelot/blob/master/ocelot/cpbd/coord_transform.py#L57) is the corresponding Ocelot function, but may want to use a different name.)
> 
> I have no opinion regarding the `xp`, `x_y_z_px_py_pz` or `xyz_pxpypz` question. But I would have the function just `to_xyz_pxpypz` (or any of the three options) because in the end it would be used as
> 
> ```python
> moments = cheetah_beam.to_xyz_pxpypz()
> another_cheetah_beam = ParameterBeam.from_xyz_pxpypz(moments, ...)
> ```

`to_xyz_pxpypz` and `from_xyz_pxpypz` sounds good

> > Regarding plotting: could we simply plot a vertical line (as if the space-charge kick was a thin element), instead of a rectangle?
> 
> A vertical line might actually be more intuitive, yes. Still, can someone explain why we do
> 
> ```python
> segment = cheetah.Segment(
>     elements=[
>         cheetah.Quadrupole(quad_length / 6),
>         cheetah.SpaceChargeKick(quad_length / 3),
>         cheetah.Quadrupole(quad_length / 3),
>         cheetah.SpaceChargeKick(quad_length / 3),
>         cheetah.Quadrupole(quad_length / 3),
>         cheetah.SpaceChargeKick(quad_length / 3),
>         cheetah.Quadrupole(quad_length / 6),
>     ]
> )
> ```
> 
> rather than
> 
> ```python
> segment = cheetah.Segment(
>     elements=[
>         cheetah.Quadrupole(quad_length / 3),
>         cheetah.SpaceChargeKick(quad_length / 3),
>         cheetah.Quadrupole(quad_length / 3),
>         cheetah.SpaceChargeKick(quad_length / 3),
>         cheetah.Quadrupole(quad_length / 3),
>         cheetah.SpaceChargeKick(quad_length / 3),
>     ]
> )
> ```
> 
> ?

I believe this is just to make the SC apply in the middle of the split drifts. They way you suggested would effectly only apply SC kicks 2 times as the last one is at the end of the tracking. In the limit of large n_split they would be the same

jank324:
> I believe this is just to make the SC apply in the middle of the split drifts. They way you suggested would effectly only apply SC kicks 2 times as the last one is at the end of the tracking. In the limit of large n_split they would be the same

Ah ... so what about the following?

```python
segment = cheetah.Segment(
    elements=[
        cheetah.SpaceChargeKick(quad_length / 3),
        cheetah.Quadrupole(quad_length / 3),
        cheetah.SpaceChargeKick(quad_length / 3),
        cheetah.Quadrupole(quad_length / 3),
        cheetah.SpaceChargeKick(quad_length / 3),
        cheetah.Quadrupole(quad_length / 3),
    ]
)
```

What I would just like to understand is that, if I place a `SpaceChargeKick` at position `s` with `effect_length = l`, it is as if space charge is considered on what interval? `[s - l, s]`, `[s, s + l]` or `[s - l / 2, s + l / 2]`? Or does it not matter? Does the question even make sense?

cr-xu:
> What I would just like to understand is that, if I place a `SpaceChargeKick` at position `s` with `effect_length = l`, it is as if space charge is considered on what interval? `[s - l, s]`, `[s, s + l]` or `[s - l / 2, s + l / 2]`? Or does it not matter? Does the question even make sense?

Hmm, I would say it doesn't make sense for small $l$, but makes a difference when the split is coarse. In the end $R_{sc}*R_{drift} \neq R_{sc}*R_{drift}$ because the transformations are noncommutative. But in the limit of $l \rightarrow 0$ then the differences become negligible. 
Let it happen in the middle ([s-2/l, s+2/l]) is probably just a better approximation (like for numerical integration)?

jank324:
> > What I would just like to understand is that, if I place a `SpaceChargeKick` at position `s` with `effect_length = l`, it is as if space charge is considered on what interval? `[s - l, s]`, `[s, s + l]` or `[s - l / 2, s + l / 2]`? Or does it not matter? Does the question even make sense?
> 
> Hmm, I would say it doesn't make sense for small l, but makes a difference when the split is coarse. In the end Rsc∗Rdrift≠Rsc∗Rdrift because the transformations are noncommutative. But in the limit of l→0 then the differences become negligible. Let it happen in the middle ([s-2/l, s+2/l]) is probably just a better approximation (similar as for numerical integration)?
> 
> Example: a beam initially with px=0 would have px=p after SC effect, so it would be more accurate to track half the drift with xp=0 and half with px=p ( `[Drift(0.5l), SC_kick(l), Drift(0.5l)]`) than the other two options:
> 
> * track the whole section with xp=0  ( `[Drift(l), SC_kick(l)]`), and
> * track the whole section with xp=p ( `[SC_kick(l), Drift(l)]`)

I think I get it now. Thanks!

jank324:
> Regarding plotting: could we simply plot a vertical line (as if the space-charge kick was a thin element), instead of a rectangle?

Done! It looks like this now:

Screenshot 2024-06-20 at 15 15 18

cheetah/accelerator/space_charge_kick.py
```diff
@@ -0,0 +1,624 @@
...
+        super().__init__(name=name)
+
+        self.effect_length = torch.as_tensor(effect_length, **self.factory_kwargs)
+        self.grid_shape = (
+            int(num_grid_points_x),
+            int(num_grid_points_y),
+            int(num_grid_points_s),
+        )
+        self.grid_extend_x = torch.as_tensor(grid_extend_x, **self.factory_kwargs)
+        # In multiples of sigma
```
jank324, line 74: Does anyone understand this comment? And shouldn't it be one line up?

cr-xu: Yes, I believe it means `grid_extend_x` is a dimensionless factor, so e.g. the `grid_size = grid_extend_x * sigma_x`

jank324:
If I didn't miss anything, this PR should be done. 🎉 I've requested final reviews from everyone and then we can merge.

jank324 (approved):

cheetah/particles/particle_beam.py
```diff
@@ -717,6 +721,72 @@ def transformed_to(
...
+        beam.particles[..., 3] = moments[..., 3] / p0.unsqueeze(-1)
+        beam.particles[..., 4] = -moments[..., 4] / beam.relativistic_beta.unsqueeze(-1)
+        beam.particles[..., 5] = (gamma - beam.relativistic_gamma.unsqueeze(-1)) / (
+            (beam.relativistic_beta * beam.relativistic_gamma).unsqueeze(-1)
+        )
+
+        return beam
+
+    def to_xyz_pxpypz(self) -> torch.Tensor:
+        """Moments in SI units as converted from the beam's `particles`."""
```
cr-xu, line 763:
```suggestion
        """Moments in SI units as converted from the beam's `particles`.
            Returns the moments tensor with shape (..., n_particles, 7)
            For each particle, the moment vector is $(x, p_x, y, p_y, z, p_z, 1)$
        """
```

tests/test_space_charge_kick.py
```diff
@@ -0,0 +1,263 @@
...
+        energy=energy,
+        radius_x=R0,
+        radius_y=R0,
+        radius_s=R0 / gamma,  # Radius of the beam in s direction in the lab frame
+        sigma_xp=torch.tensor([1e-15]),
+        sigma_yp=torch.tensor([1e-15]),
+        sigma_p=torch.tensor([1e-15]),
+    )
+
+    # Compute section lenght
```
cr-xu, line 46:
```suggestion
    # Compute section length
```

tests/test_space_charge_kick.py
```diff
@@ -0,0 +1,263 @@
...
+        energy=energy,
+        radius_x=R0,
+        radius_y=R0,
+        radius_s=R0 / gamma,  # Radius of the beam in s direction in the lab frame
+        sigma_xp=torch.tensor([1e-15]),
+        sigma_yp=torch.tensor([1e-15]),
+        sigma_p=torch.tensor([1e-15]),
+    ).broadcast(shape=(2, 3))
+
+    # Compute section lenght
```
cr-xu, line 152:
```suggestion
    # Compute section length
```

cheetah/particles/particle_beam.py
```diff
@@ -717,6 +721,75 @@ def transformed_to(
...
+            (beam.relativistic_beta * beam.relativistic_gamma).unsqueeze(-1)
+        )
+
+        return beam
+
+    def to_xyz_pxpypz(self) -> torch.Tensor:
+        """Moments in SI units as converted from the beam's `particles`.
+            Returns the moments tensor with shape (..., n_particles, 7)
+            For each particle, the moment vector is $(x, p_x, y, p_y, z, p_z, 1)$
+        """
```
cr-xu, line 766:
```suggestion
        """Moments in SI units as converted from the beam's `particles`.
        Returns the moments tensor with shape (..., n_particles, 7)
        For each particle, the moment vector is $(x, p_x, y, p_y, z, p_z, 1)$
        """
```

jank324 (approved):

RemiLehe:
> > > Regarding a replacement for `moments`, here are a few suggestions:
> > > 
> > > * `cheetah_coords_to_xp`
> > > * `cheetah_coords_to_x_y_z_px_py_pz`
> > > * `canonical_coords_to_xyz_pxpypz`
> > >   Any preferences? (For reference, [here]([https://github.com/ocelot->](https://github.com/ocelot-%3E) collab/ocelot/blob/master/ocelot/cpbd/coord_transform.py#L57) is the corresponding Ocelot function, but may want to use a different name.)
> > 
> > 
> > I have no opinion regarding the `xp`, `x_y_z_px_py_pz` or `xyz_pxpypz` question. But I would have the function just `to_xyz_pxpypz` (or any of the three options) because in the end it would be used as
> > ```python
> > moments = cheetah_beam.to_xyz_pxpypz()
> > another_cheetah_beam = ParameterBeam.from_xyz_pxpypz(moments, ...)
> > ```
> 
> `to_xyz_pxpypz` and `from_xyz_pxpypz` sounds good
> 
> > > Regarding plotting: could we simply plot a vertical line (as if the space-charge kick was a thin element), instead of a rectangle?
> > 
> > 
> > A vertical line might actually be more intuitive, yes. Still, can someone explain why we do
> > ```python
> > segment = cheetah.Segment(
> >    elements=[
> >        cheetah.Quadrupole(quad_length / 6),
> >        cheetah.SpaceChargeKick(quad_length / 3),
> >        cheetah.Quadrupole(quad_length / 3),
> >        cheetah.SpaceChargeKick(quad_length / 3),
> >        cheetah.Quadrupole(quad_length / 3),
> >        cheetah.SpaceChargeKick(quad_length / 3),
> >        cheetah.Quadrupole(quad_length / 6),
> >    ]
> > )
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > rather than
> > ```python
> > segment = cheetah.Segment(
> >    elements=[
> >        cheetah.Quadrupole(quad_length / 3),
> >        cheetah.SpaceChargeKick(quad_length / 3),
> >        cheetah.Quadrupole(quad_length / 3),
> >        cheetah.SpaceChargeKick(quad_length / 3),
> >        cheetah.Quadrupole(quad_length / 3),
> >        cheetah.SpaceChargeKick(quad_length / 3),
> >    ]
> > )
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ?
> 
> I believe this is just to make the SC apply in the middle of the split drifts. They way you suggested would effectly only apply SC kicks 2 times as the last one is at the end of the tracking. In the limit of large n_split they would be the same

@cr-xu Yes, that is correct. The way in which we apply the space-charge kicks are applied can be seen as an instance of [Strang splitting](https://en.wikipedia.org/wiki/Strang_splitting), which under reasonable conditions, is second-order accurate which respect to the length of the splitting intervals.

cheetah/accelerator/space_charge_kick.py
```diff
@@ -0,0 +1,643 @@
...
+class SpaceChargeKick(Element):
+    """
+    Applies the effect of space charge over a length `effect_length`, on the
+    **momentum** (i.e. divergence and energy spread) of the beam. The positions are
+    unmodified; this is meant to be combined with another lattice element (e.g. `Drift`)
+    that does modify the positions, but does not take into account space charge. The
+    integrated Green function method
+    (https://journals.aps.org/prab/abstract/10.1103/PhysRevSTAB.9.044204) is used to
+    compute the effect of space charge. This is similar to the method used in Ocelot.
+    The main difference is that it solves the Poisson equation in the beam frame, while
```
RemiLehe, line 27:
```suggestion
    The main difference is that Ocelot solves the Poisson equation in the beam frame, while
```

cheetah/accelerator/space_charge_kick.py
```diff
@@ -0,0 +1,643 @@
...
+        return green_func_values
+
+    def _solve_poisson_equation(
+        self, beam: ParticleBeam, moments: torch.Tensor, cell_size, grid_dimensions
+    ) -> torch.Tensor:  # Works only for ParticleBeam at this stage
+        """
+        Solves the Poisson equation for the given charge density, using FFT convolution.
+        """
+        charge_density = self._array_rho(beam, moments, cell_size, grid_dimensions)
+        charge_density_ft = torch.fft.fftn(charge_density, dim=[1, 2, 3])
```
RemiLehe, line 365: As a follow-up PR, we could use R2C and C2R FFTs, for higher efficiency.

RemiLehe: This requires modifying the shape of the arrays in Fourier space.

cheetah/accelerator/space_charge_kick.py
```diff
@@ -0,0 +1,643 @@
...
+                self.grid_extend_y * beam.sigma_y,
+                self.grid_extend_s * beam.sigma_s,
+            ],
+            dim=-1,
+        )
+
+    def _deposit_charge_on_grid(
+        self,
+        beam: ParticleBeam,
+        moments: torch.Tensor,
```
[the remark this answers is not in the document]
RemiLehe: Done.

cheetah/accelerator/space_charge_kick.py
RemiLehe: Done. In the end, it was easier than I thought.

jank324:
So, it turns out scipy `1.14.0` requires Python`>=3.10`. That's why I wasn't getting the new version. So if GitHub Actions actually runs the `3.9` test process, it should actually pass.

I guess this one early sign that `3.9` is losing support. NumPy actually recommends only supporting `3.10+` starting 5 April 2024. I would keep `3.9` in for now though. Last I checked DOOCS was only compatible with `3.9`.

cheetah/accelerator/space_charge_kick.py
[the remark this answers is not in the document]
jank324: @RemiLehe this will probably also have to be done in `particle_beam.py`, including maybe a rephrasing of the docstrings (?).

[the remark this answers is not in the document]
RemiLehe: Sounds good. Done in https://github.com/desy-ml/cheetah/pull/142/commits/b2e45c071fddd3c1a08b5da03c4eafc7876cce95

RemiLehe (approved):

ax3l:
🎉

Issue #145, opened 2024-04-26
Need more distribution options for bunch generation


cr-xu:
Currently Cheetah only supports generation of Gaussian bunches. 

More distributions should be supported, especially allowing to specify different distributions in transversal and longitudinal plane.

For example

- Truncated gaussian bunch
- Radial uniform + longitudinal uniform (cynlindrical)
- Uniform ellipsoidal (3D uniform?)

This is also requested by @RemiLehe

jp-ga:
Take a look at https://github.com/ColwynGulliford/distgen
We should implement distgen wrappers in Cheetah at some point

cr-xu:
> Take a look at https://github.com/ColwynGulliford/distgen We should implement distgen wrappers in Cheetah at some point

That looks really nice. 
So in general we should also implement an converter to import from GPT / OpenPMD beam types.

jank324:
Something else to keep in mind is that generators implemented directly in Cheetah are also differentiable ... in case someone needs that as a feature.

jp-ga:
OpenPMD already has the functionality to read and write Astra format: https://github.com/ChristopherMayes/openPMD-beamphysics/blob/master/docs/examples/write_examples.ipynb

jp-ga:
> Something else to keep in mind is that generators implemented directly in Cheetah are also differentiable ... in case someone needs that as a feature.

That's a good point.

jank324:
There is a dedicated issue for adding distgen to Cheetah now: #149

PR #146, opened 2024-04-29
Add particle generation for 3D uniform ellispoid distribution

cr-xu:
## Description

Add a preliminary 3d uniform ellispoidal generation routine.

## Motivation and Context

This addresses #145 

Right now the code is lying in `utils` but we can think of moving it later.

To get a waterbag bunch, right now one can generate the distribution first and input it into `ParticleBeam`
```python
from cheetah.utils import generate_3d_uniform_ellipsoid
from cheetah.particles import ParticleBeam
import torch

r_x = torch.tensor(1e-3)
r_y = torch.tensor(1e-3)
r_z = torch.tensor(1e-5)
num_particles = torch.tensor(100_000)

particles = generate_3d_uniform_ellipsoid(num_particles, r_x, r_y, r_z)
parray = ParticleBeam(particles=particles, energy=torch.tensor([1e6], dtype=torch.float32))
```

## Types of changes

## Checklist


cr-xu:
@jank324 Any suggestions on where to put the future routines?

@RemiLehe Is that what is needed for the benchmarking? From the papers I read somehow confusing information, is the bunch matched to the lattice or simply p_x=p_y=p_z=0?

jank324:
I think this should be a `@classmethod` of `ParameterBeam` named `3d_uniform_elipsoid`, in the same way as `from_parameters` and `from_twiss`. Along that line of thinking it probably makes sense to rename and merge those two at some point into a single method called something like `3d_gaussian_elipsoid` that takes either Twiss parameters or like sigmas and such (?)

cheetah/utils.py
```diff
@@ -9,3 +12,47 @@ def __call__(self):
         name = f"{self._prefix}_{self._counter}"
         self._counter += 1
         return name
+
+
+def generate_3d_uniform_ellipsoid(
+    num_particles: torch.Tensor, r_x: torch.Tensor, r_y: torch.Tensor, r_z: torch.Tensor
```
jank324, line 18: I think the arguments should be called `radius_x` and so on. It's not too long and this way I don't have to look at the documentation to figure out what `r_x` is ... i.e. it's much more readable.

cr-xu:
> I think this should be a `@classmethod` of `ParameterBeam` named `3d_uniform_elipsoid`, in the same way as `from_parameters` and `from_twiss`. Along that line of thinking it probably makes sense to rename and merge those two at some point into a single method called something like `3d_gaussian_elipsoid` that takes either Twiss parameters or like sigmas and such (?)

Yes, that could be a solution. I don't know if it will become very crowded in the future, as particle distributions can be quite arbitrary. E.g. transverse 2D + longitudinal 1D with different distribtuions; 
Technically what we already have is also only a decoupled case. e.g. x-x', y-y', $\tau - \delta$, rather than a fully-correlated gaussian ellipsoid (with x-y, x-z, y-z,...)

RemiLehe:
Thanks for adding this!
Regarding your question: yes, we would only need a cold beam for now (zero divergence, zero energy spread)

jank324:
> > I think this should be a `@classmethod` of `ParameterBeam` named `3d_uniform_elipsoid`, in the same way as `from_parameters` and `from_twiss`. Along that line of thinking it probably makes sense to rename and merge those two at some point into a single method called something like `3d_gaussian_elipsoid` that takes either Twiss parameters or like sigmas and such (?)
> 
> Yes, that could be a solution. I don't know if it will become very crowded in the future, as particle distributions can be quite arbitrary. E.g. transverse 2D + longitudinal 1D with different distributions; Technically what we already have is also only a decoupled case. e.g. x-x', y-y', τ−δ, rather than a fully-correlated gaussian ellipsoid (with x-y, x-z, y-z,...)

I'm not too worried that it would get too crowded. I think

```python
beam = ParticleBeam.3d_uniform_elipsoid()
```

makes for a very intuitive interface. If I were to look for the distribution code, I would also immediately look where the `ParticleBeam` source is. I think it's just important, that we keep it structured in the sense that we don't interleave other `ParticleBeam` methods with the class methods for distributions.

Also, I think that at some point `Beam`, `ParameterBeam` and `ParticleBeam` might end up in separate files.

jank324 (approved):
@cr-xu I reviewed the code by breaking and refactoring it to be more readable. I think it can be merged as it is now.

PR #148, opened 2024-04-30
Update pre-commit config

### LINKED ISSUE #147: Use pre-commit instead of a standalone format Gtihub action in CI/CD
This format pipeline failed even with local pre-commit hooks
https://github.com/desy-ml/cheetah/actions/runs/8876067584/job/24366827967

I think the reason is a version conflict in black (22.x -> 24.x)

So to make things easier in the future, there are different ways to proceed:
1. Regularly update the `.pre-commit-config.yaml` so that the versions are compatible with the CI/CD workflow 
2. Another way around is to fix the formatting tools' version to the ones in `.pre-commit-config.yaml`
3. Replace the format action with a pre-commit action which does formatting check and refomat things automatically if possible.

jank324:
## Description

Update alle versions in `.pre-commit-config.yaml`.

## Motivation and Context

Closes #147.

## Types of changes

## Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,19 +1,19 @@
 repos:
--   repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v2.3.0
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.6.0
     hooks:
-    -   id: check-yaml
--   repo: https://github.com/psf/black
-    rev: 22.10.0
+      - id: check-yaml
+  - repo: https://github.com/psf/black
+    rev: 24.4.2
     hooks:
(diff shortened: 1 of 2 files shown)


jank324:
So @cr-xu, I'm not sure if this is a satisfactory solution. Basically, the versions of `rev` are not allowed to be dynamic. For now, I've updated them. We would have to keep updating them whenever a version mismatch leads to a mismatch between the pre-commit hook and the respective GitHub Action. This would be a little work ... but the mismatch would inherently alert us when it needs doing.

cr-xu:
okay it makes sense. I think breaking changes in these linting tools happen only more or less on a yearly basis so it wouldn't be so annoying.

PR #153, opened 2024-05-08
add `BmadQuadrupole` element

### LINKED ISSUE #139: Add quadrupole with chromatic effects
Add quadrupole with chromatic effects in Cheetah. It will need a coordinate transform between ocelot and canonical phase space coordinates.

jp-ga:
## Description

Add `BmadQuadrupole` element by interfacing with Bmad-X tracking routines. 

## Motivation and Context

Closes #139 

## Types of changes

## Checklist


cheetah/accelerator.py
```diff
@@ -2442,3 +2442,143 @@ def __repr__(self) -> str:
...
+    :param name: Unique identifier of the element.
+    """
+
+    def __init__(
+        self,
+        length: Union[torch.Tensor, nn.Parameter],
+        k1: Optional[Union[torch.Tensor, nn.Parameter]] = None,
+        misalignment: Optional[Union[torch.Tensor, nn.Parameter]] = None,
+        tilt: Optional[Union[torch.Tensor, nn.Parameter]] = None,
+        num_steps: int = 10,
```
jank324, line 2468: Is this the only argument that's different between a Cheetah quad and a Bmad-X quad?

cheetah/bmadx/interface.py
```diff
@@ -0,0 +1,166 @@
...
+        dtype = bmadx_particle.x.dtype
+    )
+
+    return cheetah_beam
+
+
+def cheetah_to_bmad_quad(cheetah_quad) -> Quadrupole :
+    """Converts a Cheetah Quadrupole to a BmadX Quadrupole."""
+    bmadx_quad = Quadrupole(
+        L = cheetah_quad.length,
```
jank324, line 147: There seem to still be some black errors. This line is one of them.

jank324:
As a general comment. I'm not sure I like this quick-and-dirty drop-in way of integrating Bmad-X elements. However, I think it makes sense to test them this way. In the long-term we should think about how this can be clean up better.

Some questions we should ask ourselves:
 - Do we really want a separate `bmadx` module? This means that code is suddenly in a very different place than the other tracking code.
 - Should there be a single `Quadrupole` class that has something like a `tracking_method: Literal["linear", "bmadx"]` argument?
 
Maybe slightly unrelated but I saw it in this PR:
 - I think it's not very nice to have something like `C = scipy.constants.speed_of_light` at the top of a file. Ideally, one would just `from scripy.constants import speed_of_light` and use that. I think this might cause issues sometimes, when a `torch.Tensor` is strictly needed, but I cannot produce an example right now. Maybe we should add a `cheetah.constants` module that takes `scripy.constnts` and always returns `torch.Tensor`s. But again, I'm not sure we actually need that ... just something to think about.

jp-ga:
One small thing to keep in mind, bmadx tracking agrees to single precision with bmadx. I was expecting it to agree to double precision, but it is good enough for now.

cheetah/bmadx_utils.py
```diff
@@ -0,0 +1,148 @@
...
+    # Bmad coords:
+    bmad_coords[..., 4] = z
+    bmad_coords[..., 5] = pz
+
+    return bmad_coords, p0c
+
+
+def bmad_to_cheetah_coords(
+    bmad_coords: torch.Tensor, p0c: torch.Tensor, mc2: torch.Tensor
+) -> torch.Tensor:
```
cr-xu, line 35: Would it make sense to just integrate these two methods also in `ParticleBeam` like the other transformations?
I think `p0c` can also be provided as a property.

cheetah/bmadx_utils.py
```diff
@@ -0,0 +1,148 @@
...
+    delta = (energy - ref_energy) / p0c
+
+    # Cheetah coords:
+    cheetah_coords[..., 4] = tau
+    cheetah_coords[..., 5] = delta
+
+    return cheetah_coords, ref_energy
+
+
+def offset_particle_set(x_offset, y_offset, tilt, x_lab, px_lab, y_lab, py_lab):
```
cr-xu, line 65: Here is missing some docstring and variable typing

cheetah/bmadx_utils.py
```diff
@@ -0,0 +1,148 @@
...
+    x_lab_int = x_ele * c - y_ele * s
+    y_lab_int = x_ele * s + y_ele * c
+    x_lab = x_lab_int + x_offset
+    y_lab = y_lab_int + y_offset
+    px_lab = px_ele * c - py_ele * s
+    py_lab = px_ele * s + py_ele * c
+    return x_lab, px_lab, y_lab, py_lab
+
+
+def low_energy_z_correction(pz, p0c, mc2, ds):
```
cr-xu, line 89:
```suggestion
def low_energy_z_correction(pz: torch.Tensor, p0c: torch.Tensor, mc2: torch.Tensor, ds: torch.Tensor) -> torch.Tensor:
```

cheetah/accelerator/quadrupole.py
```diff
@@ -72,6 +93,135 @@ def transfer_map(self, energy: torch.Tensor) -> torch.Tensor:
...
+        b1 = k1 * length
+
+        # Begin bmadx tracking:
+        x, px, y, py = offset_particle_set(x_off, y_off, tilt, x, px, y, py)
+
+        for _i in range(n_step):
+            rel_p = 1 + pz  # Particle's relative momentum (P/P0)
+            k1 = b1 / (length * rel_p)
+
+            tx, dzx = quad_mat2_calc(-k1, step_len, rel_p)
```
cr-xu, line 179: Maybe rename the `quad_mat2_calc` to something more readable like `calculate_second_order_quadrupole_matrix`?

jp-ga: renamed

cheetah/bmadx_utils.py
jp-ga: done

cheetah/bmadx_utils.py
jp-ga: done

jank324:
@jp-ga, the question came up with @roussel-ryan, if this or #163 should be merged first. I think a main question in that was if the changes in #163 might make conversions in this PR redundant. Do you have an opinion on this?

cheetah/bmadx_utils.py
jp-ga: I was thinking to include the coordinate transforms in here instead: https://github.com/desy-ml/cheetah/blob/master/cheetah/converters/bmad.py

This was it could also be used to transform beams for both Bmad and Bmad-X. 

Any thoughts?

jp-ga:
> @jp-ga, the question came up with @roussel-ryan, if this or #163 should be merged first. I think a main question in that was if the changes in #163 might make conversions in this PR redundant. Do you have an opinion on this?

The transformations in this PR assume the canonical coordinates as describe in #163, so you can merge #163 first if needed.

tests/test_quadrupole.py
```diff
@@ -124,3 +124,35 @@ def test_tilted_quadrupole_multiple_batch_dimension():
     outgoing = segment(incoming)
 
     assert torch.allclose(outgoing.particles[0, 0], outgoing.particles[0, 1])
+
+
+def test_quadrupole_bmadx_tracking():
+    incoming_beam = torch.load("tests/resources/quadrupole_bmadx/incoming_beam.pt")
```
roussel-ryan, line 130: @jp-ga I just ran into some vectorization issues when specifying multiple quadrupole strengths and then tracking (as we do in GPSR), can you add a test for this if its not elsewhere?

jank324: @jp-ga has this been fixed?

jank324:
I'm not sure if we should add now how `Quadrupole.split` handles the `num_steps` argument.

jank324 (changes requested):
@jp-ga Can you have a brief look at the `split` function I wrote a comment about?

@roussel-ryan Can you check if the test I added seems to be about the error you ran into?

Otherwise, I think this is ready to merge!

jp-ga:
> I'm not sure if we should add now how `Quadrupole.split` handles the `num_steps` argument.

@jank324 I thought about this back when developing Bmad-X, and at the end I decided that it was better to keep the number of steps in the quad different and independent from the number of elements when splitting. For example, if you have a quad with num_steps=3 and you split the element in 5, the code should do 5 quadrupoles with L/5 length, each of which with num_steps=3 (doing 3 drift-kick-drift steps). 

In the future I would like to implement fringe fields, so the splitting should have entrance fringe at the first element's entrance and exit fringe at the last element's exit, but we can leave this for a future PR.

jank324:
> > I'm not sure if we should add now how `Quadrupole.split` handles the `num_steps` argument.
> 
> @jank324 I thought about this back when developing Bmad-X, and at the end I decided that it was better to keep the number of steps in the quad different and independent from the number of elements when splitting. For example, if you have a quad with num_steps=3 and you split the element in 5, the code should do 5 quadrupoles with L/5 length, each of which with num_steps=3 (doing 3 drift-kick-drift steps).
> 
> In the future I would like to implement fringe fields, so the splitting should have entrance fringe at the first element's entrance and exit fringe at the last element's exit, but we can leave this for a future PR.

So the behaviour you would like to have now is that each of the 5 child quadrupoles has 3 splits, just like their parent had 3 splits? In that case, don't we need to add the line highlighted below in the `split` method?

```python
element = Quadrupole(
    torch.min(resolution, remaining),
    self.k1,
    misalignment=self.misalignment,
    tilt=self.tilt,
    num_steps=self.num_steps,   # <-- Add this line?
    dtype=self.length.dtype,
    device=self.length.device,
)
```

roussel-ryan:
@jank324 I'm not sure if the test you added will raise the issue I ran into, I think my error had to do with the ellipsoid beam generation, but I'm not sure if that test will run into the same issue

jp-ga:
> > > I'm not sure if we should add now how `Quadrupole.split` handles the `num_steps` argument.
> > 
> > 
> > @jank324 I thought about this back when developing Bmad-X, and at the end I decided that it was better to keep the number of steps in the quad different and independent from the number of elements when splitting. For example, if you have a quad with num_steps=3 and you split the element in 5, the code should do 5 quadrupoles with L/5 length, each of which with num_steps=3 (doing 3 drift-kick-drift steps).
> > In the future I would like to implement fringe fields, so the splitting should have entrance fringe at the first element's entrance and exit fringe at the last element's exit, but we can leave this for a future PR.
> 
> @jp-ga So the behaviour you would like to have now is that each of the 5 child quadrupoles has 3 splits, just like their parent had 3 splits? In that case, don't we need to add the line highlighted below in the `split` method?
> 
> ```python
> element = Quadrupole(
>     torch.min(resolution, remaining),
>     self.k1,
>     misalignment=self.misalignment,
>     tilt=self.tilt,
>     num_steps=self.num_steps,   # <-- Add this line?
>     dtype=self.length.dtype,
>     device=self.length.device,
> )
> ```

@jank324 Exactly, I will add it soon.

jank324 (approved):

cheetah/utils/bmadx.py
```diff
@@ -0,0 +1,213 @@
...
+    :param length: Quadrupole length.
+    :param rel_p: Relative momentum P/P0.
+    :return: Tuple of transfer matrix elements and coefficients.
+        a11, a12, a21, a22: Transfer matrix elements.
+        c1, c2, c3: Second order derivatives of z such that
+            z = c1 * x_0^2 + c2 * x_0 * px_0 + c3 * px_0^2.
+    """
+    # TODO: Revisit to fix accumulated error due to machine epsilon
+
+    eps = 2.220446049250313e-16  # Machine epsilon to double precission
```
cr-xu, line 194:
```suggestion
def calculate_quadrupole_coefficients(
    k1: torch.Tensor,
    length: torch.Tensor,
    rel_p: torch.Tensor,
    eps: float=np.finfo(np.float64).eps,
) -> tuple[list[torch.Tensor], list[torch.Tensor]]:
    """
    Returns 2x2 transfer matrix elements aij and the coefficients to calculate the
    change in z position.

    NOTE: Accumulated error due to machine epsilon.

    :param k1: Quadrupole strength (k1 > 0 ==> defocus).
    :param length: Quadrupole length.
    :param rel_p: Relative momentum P/P0.
    :param eps: Machine precision epsilon, default to double precision.
    :return: Tuple of transfer matrix elements and coefficients.
        a11, a12, a21, a22: Transfer matrix elements.
        c1, c2, c3: Second order derivatives of z such that
            z = c1 * x_0^2 + c2 * x_0 * px_0 + c3 * px_0^2.
    """
    # TODO: Revisit to fix accumulated error due to machine epsilon

```

cheetah/utils/bmadx.py
```diff
@@ -0,0 +1,213 @@
+import torch
```
cr-xu, line 1:
```suggestion
import numpy as np
import torch
```

cheetah/utils/bmadx.py
```diff
@@ -0,0 +1,217 @@
...
+    )
+
+    return dz
+
+
+def calculate_quadrupole_coefficients(
+    k1: torch.Tensor,
+    length: torch.Tensor,
+    rel_p: torch.Tensor,
+    eps: float=np.finfo(np.float64).eps,
```
cr-xu, line 180:
```suggestion
    eps: float = np.finfo(np.float64).eps,
```

cheetah/utils/bmadx.py
```diff
@@ -0,0 +1,217 @@
...
+    :param rel_p: Relative momentum P/P0.
+    :param eps: Machine precision epsilon, default to double precision.
+    :return: Tuple of transfer matrix elements and coefficients.
+        a11, a12, a21, a22: Transfer matrix elements.
+        c1, c2, c3: Second order derivatives of z such that
+            z = c1 * x_0^2 + c2 * x_0 * px_0 + c3 * px_0^2.
+    """
+    # TODO: Revisit to fix accumulated error due to machine epsilon
+
+
```
cr-xu, line 199:
```suggestion

```

cheetah/utils/bmadx.py
```diff
@@ -0,0 +1,215 @@
+import numpy as np
+import torch
+
+
```
cr-xu, line 4:
```suggestion
double_precision_epsilon = np.finfo(np.float64).eps

```

cheetah/utils/bmadx.py
```diff
@@ -0,0 +1,215 @@
...
+    )
+
+    return dz
+
+
+def calculate_quadrupole_coefficients(
+    k1: torch.Tensor,
+    length: torch.Tensor,
+    rel_p: torch.Tensor,
+    eps: float = np.finfo(np.float64).eps,
```
cr-xu, line 180:
```suggestion
    eps: float = double_precision_epsilon,
```

cheetah/utils/bmadx.py
```diff
@@ -0,0 +1,216 @@
+import numpy as np
+import torch
+
+double_precision_epsilon = np.finfo(np.float64).eps
+
+def cheetah_to_bmad_coords(
```
cr-xu, line 6:
```suggestion
double_precision_epsilon = np.finfo(np.float64).eps


def cheetah_to_bmad_coords(
```

cheetah/utils/bmadx.py
```diff
@@ -1,5 +1,8 @@
+import numpy as np
 import torch
 
+double_precision_epsilon = np.finfo(np.float64).eps
```
jank324, line 4: Is there no `torch` equivalent for this? I would also just pass this directly as the default argument rather than create a variable for it.

cr-xu: I tried that but flake8 doesn't like passing directly a function call to default argument

cheetah/utils/bmadx.py
```diff
@@ -0,0 +1,217 @@
+import numpy as np
+import torch
+
+double_precision_epsilon = np.finfo(np.float64).eps
```
cr-xu, line 4:
```suggestion
import torch

double_precision_epsilon = torch.finfo(torch.float64).eps
```

cheetah/utils/bmadx.py
jank324: Ah ... I see ... this is probably not an issue here, but let's stay flake8-conform then and keep the variable.

cr-xu (approved):
One final comment: Is it intended to keep the `dev/` folder? If the notebook is useful, shall we move it to some other places, like a documentation page?
Otherwise it's good to go!

jank324:
> One final comment: Is it intended to keep the `dev/` folder? If the notebook is useful, shall we move it to some other places, like a documentation page? Otherwise it's good to go!

I'm also not sure about the `dev/` folder. I left it because I also sometimes write notebooks to develop features and I often worry that I will have to write them twice because I always delete them at some point once I'm done and I never commit them. `dev/` would be a clean way to keep them around.

On the other hand, I've never actually felt like I needed to rewrite a notebook for some reason and we might not want to clutter up the repository and history like this.

jank324:
So ... to document the decision: The `dev/` folder remains for now, but the long-term goal should be to not have it. Interesting examples should instead, for example, appear in the documentation.

PR #200, opened 2024-06-27
Use KDE for beam screen images

cr-xu:
## Description

Add an option for `Screen` element. Now it can use KDE to get the beam image which allows differentiation.
Behavior can be switched using `method="histogram"` and `method="kde"`.
It contains a new feature `kde_bandwidth` for smoothing the beam image.

The returned image has the convention [...,Y, X] where y-axis starts from the top of the image, e.g. `plt.imshow()` will show the image upright

```python

import cheetah
import torch
import matplotlib.pyplot as plt

incoming = cheetah.ParticleBeam.from_parameters(num_particles = 10000, mu_x = torch.tensor([-0.5e-3]), mu_y=torch.tensor([1.0e-3]), sigma_x = torch.tensor([1e-4]), sigma_y = torch.tensor([3e-4]))

screen = cheetah.Screen(resolution=(1200, 800), pixel_size=(5e-6, 5e-6), binning=torch.tensor(2), is_active=True)

screen.track(incoming)

plt.imshow(screen.reading[0], extent=screen.extent, cmap='gray')
plt.xlabel('x [m]')
plt.ylabel('y [m]')
```

image

## Motivation and Context

## Types of changes

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 - Add `name` to `Drift` element `__repr__` (see #201) (@ansantam)
 - Fix bug where `dtype` was not used when creating a `ParameterBeam` from Twiss parameters (see #206) (@jank324)
 - Fix bug after running `Segment.inactive_elements_as_drifts` the drifts could have the wrong `dtype` (see #206) (@jank324)
+- Fix an issue where splitting elements would result in splits with a different `dtype` (see #211) (@jank324)
 
 ### 🐆 Other
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 - Implement space charge using Green's function in a `SpaceChargeKick` element (see #142) (@greglenerd, @RemiLehe, @ax3l, @cr-xu, @jank324)
 - `Segment`s can now be imported from Bmad to devices other than `torch.device("cpu")` and dtypes other than `torch.float32` (see #196, #206) (@jank324)
 - `Screen` will now use KDE for differentiable images. (see #200) (@cr-xu, @roussel-ryan)
+- Moving `Element`s and `Beam`s to a different `device` and changing their `dtype` like with any `torch.nn.Module` is now possible (see #209) (@jank324)
 
 ### 🐛 Bug fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
 - Add a new class method for `ParticleBeam` to generate a 3D uniformly distributed ellipsoidal beam (see #146) (@cr-xu, @jank324)
 - Add Python 3.12 support (see #161) (@jank324)
 - Implement space charge using Green's function in a `SpaceChargeKick` element (see #142) (@greglenerd, @RemiLehe, @ax3l, @cr-xu, @jank324)
-- `Segment`s can now be imported from Bmad to devices other than `torch.device("cpu")` (see #196) (@jank324)
+- `Segment`s can now be imported from Bmad to devices other than `torch.device("cpu")` and dtypes other than `torch.float32` (see #196, #206) (@jank324)
 - `Screen` will now use KDE for differentiable images. (see #200) (@cr-xu, @roussel-ryan)
 
 ### 🐛 Bug fixes
@@ -20,6 +20,9 @@
 - Now all `Element` have a default length of `torch.zeros((1))`, fixing occasional issues with using elements without length, such as `Marker`, `BPM`, `Screen`, and `Aperture`. (see #143) (@cr-xu)
 - Fix bug in `Cavity` `_track_beam` (see #150) (@jp-ga)
 - Fix issue where dipoles would not get a unique name by default (see #186) (@hespe)
+- Add `name` to `Drift` element `__repr__` (see #201) (@ansantam)
+- Fix bug where `dtype` was not used when creating a `ParameterBeam` from Twiss parameters (see #206) (@jank324)
+- Fix bug after running `Segment.inactive_elements_as_drifts` the drifts could have the wrong `dtype` (see #206) (@jank324)
 
 ### 🐆 Other
 
@@ -30,6 +33,7 @@
 - Rename converter modules to the respective name of the accelerator code (see #167) (@jank324)
 - Added imports to the code example in the README (see #188) (@jank324)
 - Refactor definitions of physical constants (see #189) (@hespe)
+- Fix the quadrupole strength units in the quadrupole docstring (see #202) (@ansantam)
 
 ## [v0.6.3](https://github.com/desy-ml/cheetah/releases/tag/v0.6.3) (2024-03-28)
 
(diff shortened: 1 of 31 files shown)


cr-xu:
Hmm... so KDE clearly makes the screen reading slower, as expected :)

@jank324 Shall we 
- Change the test file to make the speed limit more generous?
- Still provide a basic histogramdd() reading for normal tracking, when differentiation w.r.t. images are not needed? This can be switched e.g. with `use_kde=False` by default?

jank324:
I think we should provide both. Im just wondering whether it should be `use_kde=True` or `method=`kde`.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 - Add Python 3.12 support (see #161) (@jank324)
 - Implement space charge using Green's function in a `SpaceChargeKick` element (see #142) (@greglenerd, @RemiLehe, @ax3l, @cr-xu, @jank324)
 - `Segment`s can now be imported from Bmad to devices other than `torch.device("cpu")` and dtypes other than `torch.float32` (see #196, #206) (@jank324)
+- `Screen` will now use KDE for differentiable images. (see #200) (@cr-xu, @roussel-ryan
```
jank324, line 16:
```suggestion
- `Screen` will now use KDE for differentiable images. (see #200) (@cr-xu, @roussel-ryan)
```

cheetah/accelerator/screen.py
```diff
@@ -38,7 +43,9 @@ def __init__(
         pixel_size: Optional[Union[torch.Tensor, nn.Parameter]] = None,
         binning: Optional[Union[torch.Tensor, nn.Parameter]] = None,
         misalignment: Optional[Union[torch.Tensor, nn.Parameter]] = None,
+        kde_bandwith: Optional[Union[torch.Tensor, nn.Parameter]] = None,
         is_active: bool = False,
+        method: str = "histogram",
```
jank324, line 48:
```suggestion
        method: Literal["histogram", "kde"] = "histogram",
```

cheetah/utils.py
```diff
@@ -9,3 +18,170 @@ def __call__(self):
...
+
+    if not isinstance(bins, torch.Tensor):
+        raise TypeError(f"Input bins type is not a torch.Tensor. Got {type(bins)}")
+
+    if not isinstance(sigma, torch.Tensor):
+        raise TypeError(f"Input sigma type is not a torch.Tensor. Got {type(sigma)}")
+
+    if not bins.dim() == 1:
+        raise ValueError(
+            "Input bins must be a of the shape NUM_BINS" " Got {}".format(bins.shape)
```
jank324, line 58: We should use f-strings.

cheetah/utils.py
```diff
@@ -9,3 +18,170 @@ def __call__(self):
...
+    Returns:
+        Computed histogram of shape :math:`(B, N_{bins}), N_{bins})`.
+
+    Examples:
+        >>> x1 = torch.rand(2, 32)
+        >>> x2 = torch.rand(2, 32)
+        >>> bins = torch.torch.linspace(0, 255, 128)
+        >>> hist = kde_histogram_2d(x1, x2, bins, bandwidth=torch.tensor(0.9))
+        >>> hist.shape
+        torch.Size([2, 128, 128])
```
jank324, line 179: I think this example doesn't actually match the modified version (?)

cheetah/utils.py
```diff
@@ -9,3 +18,170 @@ def __call__(self):
...
+    return pdf
+
+
+def kde_histogram_2d(
+    x1: torch.Tensor,
+    x2: torch.Tensor,
+    bins1: torch.Tensor,
+    bins2: torch.Tensor,
+    bandwidth: torch.Tensor,
+    weights=None,
```
jank324, line 155: What type is `weights`?

jank324:
@cr-xu could you have a look at the docstrings? Some of the returns just said what type the return is, but not what it is, and I wasn't sure.

jank324:
Also, looking at this, would it make sense to split `utils.py` into
```
utils/__init__.py
utils/kde.py
utils/unique_name_generator.py
```
within this PR or do that later on?

jank324:
And last one, @cr-xu was there a reason you chose this implementation over the one you had? Just wondering.

cr-xu:
> And last one, @cr-xu was there a reason you chose this implementation over the one you had? Just wondering.

They are equivalent in math but I find this implementation just cleaner and more readable.

> @cr-xu could you have a look at the docstrings? Some of the returns just said what type the return is, but not what it is, and I wasn't sure.

I believe you fixed all the docstrings, right? Thanks!

jank324:
> > And last one, @cr-xu was there a reason you chose this implementation over the one you had? Just wondering.
> 
> They are equivalent in math but I find this implementation just cleaner and more readable.
> 
> > @cr-xu could you have a look at the docstrings? Some of the returns just said what type the return is, but not what it is, and I wasn't sure.
> 
> I believe you fixed all the docstrings, right? Thanks!

I didn't fix every docstring ... just the ones I could easily figure out what the return is. So it would be good if you could look over if there are missing ones, and if what I put in makes sense.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 - Add Python 3.12 support (see #161) (@jank324)
 - Implement space charge using Green's function in a `SpaceChargeKick` element (see #142) (@greglenerd, @RemiLehe, @ax3l, @cr-xu, @jank324)
 - `Segment`s can now be imported from Bmad to devices other than `torch.device("cpu")` and dtypes other than `torch.float32` (see #196, #206) (@jank324)
+- `Screen` will now use KDE for differentiable images. (see #200) (@cr-xu, @roussel-ryan)
```
jank324, line 16:
```suggestion
- `Screen` now offers the option to use KDE for differentiable images (see #200) (@cr-xu, @roussel-ryan)
```

cr-xu:
> I didn't fix every docstring ... just the ones I could easily figure out what the return is. So it would be good if you could look over if there are missing ones, and if what I put in makes sense.

I looked over the code again and fixed a few things. Now the PR is good to go from my side. 
A potential TODO would be to add an example case in the documentation page, but that can be done in future.

tests/test_kde.py
```diff
@@ -0,0 +1,51 @@
...
+    )
+
+    hist_neglect_weights = kde_histogram_2d(
+        x_weighted[:, 0], x_weighted[:, 1], bins1, bins2, sigma
+    )
+    print(hist_unweighted[5])
+    print(hist_weighted[5])
+    print(hist_neglect_weights[5])
+    assert torch.allclose(hist_unweighted, hist_weighted)
+    assert not torch.allclose(hist_weighted, hist_neglect_weights)
```
jank324, line 51: @cr-xu could you add a description to these?

jank324 (approved):

cr-xu:
Here's a quick study of the computation scaling for KDE and histogram methods:

The pixel counts are the total numbers in a screen (e.g. 1e6 corresponds to 1000*1000 screen).

image

Issue #203, opened 2024-06-28
New Corrector class


ansantam:
A corrector that can kick in both planes simultaneously is missing.

One could sequentially use `HorizontalCorrector` and `VerticalCorrector`, but then the one-to-one element conversion does not work and complicates the conversion functions.

I suggest to create a new class `Corrector`, with `HorizontalCorrector` and `VerticalCorrector` as subclasses.

jank324:
This raises a question I've been pondering before, if `Steerer` might be a better name? However, that would introduce a breaking change. It might also be worth seeing what @cr-xu's opinion is in that regard?

ansantam:
In MAD-X they use the name `KICKER`, which is very descriptive but also slang.
I think `Corrector` is accurate and general enough.

jank324:
`Kicker` to me is one that can be turned on and off very quickly ... so it "kicks" 😄 I mostly use `Steerer` in papers now because they don't always "correct", but I also use the letter `C` still. I guess if it doesn't really matter that much, it's probably just best to avoid introducing breaking changes.

PR #207, opened 2024-06-28
203 new corrector class

ansantam:
## Description

Implementation of a new class `Corrector`. 
To build the `Corrector` class I took the old `HorizontalCorrector` class and added an extra entry in the transfer matrix to include `yp` (vertical angle). Now instead of the `angle` property, you have `horizontal_angle` and `vertical_angle`.
- `xp` is in position (1,6) 
- `yp` is in position (3,6)

I then derived `HorizontalCorrector` and `VerticalCorrector` from the `Corrector` class, and removed either `vertical_angle` or `horizontal_angle` as property. This is checked in the corresponding test scripts (e.g. one cannot feed the `horizontal_angle` property to `VerticalCorrector`).

One could make it in a way that `HorizontalCorrector` and `VerticalCorrector` both have the `angle` property, but I don't know how to do that and the current solution works.

## Motivation and Context

The motivation is because I need it to continue with the MAD-X converter, using the AWAKE lattice which contains combined correctors. Probably also useful to other people in the future.

Closes issue #203 

## Types of changes

## Checklist

- [] I have run `pytest` on a machine with a CUDA GPU and made sure all tests pass (**required**).

### DIFF
--- a/tests/test_horizontal_corrector.py
+++ b/tests/test_horizontal_corrector.py
@@ -1,17 +1,16 @@
 import torch
 
-from cheetah import Corrector, Drift, ParameterBeam, ParticleBeam, Segment
+from cheetah import Drift, HorizontalCorrector, ParameterBeam, ParticleBeam, Segment
 
 
-def test_corrector_off():
+def test_horizontal_corrector_off():
     """
-    Test that a corrector with horizontal_angle=0 and vertical_angle=0 behaves
-    still like a drift.
+    Test that a corrector with horizontal_angle=0 behaves
+    still like a drift and that the angle translates properly.
     """
-    corrector = Corrector(
+    corrector = HorizontalCorrector(
         length=torch.tensor([0.3]),
         horizontal_angle=torch.tensor([0.0]),
-        vertical_angle=torch.tensor([0.0]),
     )
     drift = Drift(length=torch.tensor([1.0]))
     incoming_beam = ParameterBeam.from_twiss(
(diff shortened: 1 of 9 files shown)


jank324:
So ... changing the failing test as in the following works:

```python
def test_ea_magnets():
    """
    Test that gradients are tracking when the magnet settings in the ARES experimental
    area require grad.
    """
    ea = cheetah.Segment.from_ocelot(ares.cell, warnings=False).subcell(
        "AREASOLA1", "AREABSCR1"
    )
    incoming_beam = cheetah.ParticleBeam.from_astra(
    )

    ea.AREAMQZM2.k1 = nn.Parameter(ea.AREAMQZM2.k1)
    ea.AREAMQZM1.k1 = nn.Parameter(ea.AREAMQZM1.k1)
    # ea.AREAMCVM1.angle = nn.Parameter(ea.AREAMCVM1.angle)
    ea.AREAMCVM1.horizontal_angle = nn.Parameter(ea.AREAMCVM1.horizontal_angle)
    ea.AREAMQZM3.k1 = nn.Parameter(ea.AREAMQZM3.k1)
    # ea.AREAMCHM1.angle = nn.Parameter(ea.AREAMCHM1.angle)
    ea.AREAMCHM1.horizontal_angle = nn.Parameter(ea.AREAMCHM1.horizontal_angle)

    outgoing_beam = ea.track(incoming_beam)

    assert outgoing_beam.particles.grad_fn is not None
    ```
    
 This pretty much confirms the issue. So the question is, do we solve the `Parameter` / `Buffer` assignment issue as a whole now before this can be merged, or do we merge this as a hack?
    
In terms of a hack: Just ignoring the test would in some potentially unforeseen situations break differentiability. I think a better hack would be just to make three different classes. There is even a case for that not being a hack. (Maybe a corrector transfer map function and three different classes use it (kind of like `base_rmatrix`.)

File: tests/test_horizontal_corrector.py [hunk 1]
```diff
@@ -0,0 +1,78 @@
...
+
+
+def test_horizontal_angle_property():
+    try:
+        HorizontalCorrector(
+            length=torch.tensor([0.3]),
+            vertical_angle=torch.tensor([0.0]),
+        )
+    except TypeError:
+        pass
```
jank324, line 51: @ansantam, what is this? 🦦

jank324 (changes requested):
Change to matrix generating function instead of inheritance as we talked about.

jank324:
A thought I had today: All the trouble with the subclasses is probably not worth it for such a simple element. It probably makes more sense to just implement the combined element entirely on its own.

amylizzle:
Adopted in #579

Issue #138, opened 2024-03-27
Automatic broadcasting


jank324:
In #116, we added a vectorised way of using Cheetah. This means that instead of either one of these

```python
elem1 = Drift(length=0.3142)
elem2 = Drift(length=torch.tensor(0.3142))
```

you have to remember to set properties in the correct dimension even if you are not using multiple dimensions or only a single value for that property

```python
elemt3 = Drift(length=torch.tensor([0.3142])
```

By introducing automatic broadcasting (in a similar way to how PyTorch does it), we want to be able to use either one of the examples in the first code block while tracking, for example, a beam with 1000 different properties.

cr-xu:
Is this now considered solved by #208?

jank324:
> Is this now considered solved by #208?

Yes!

PR #208, opened 2024-07-02
Enable automatic broadcasting

roussel-ryan:
## Description

Adds the ability to do automatic broadcasting according to PyTorch conventions without the use of broadcast function.

PyTorch broadcast semantics rely on the following rules
- Each tensor has at least one dimension.
- When iterating over the dimension sizes, starting at the trailing dimension, the dimension sizes must either be equal, one of them is 1, or one of them does not exist.

This PR allows the specification of arbitrary tensor sizes for beamline attributes and particle beam attributes as long as they abide by the rules above. For example,
```python
quadrupole = cheetah.Quadrupole(
    length=torch.tensor([[0.2, 0.25], [0.3, 0.35], [0.4, 0.45]]), # shape = (3,2)
    k1=torch.tensor([[4.2], [4.3], [4.4]]), # shape = (3,1) 
)
incoming = cheetah.ParticleBeam.from_parameters(
    num_particles=100_000,
    sigma_x=torch.tensor([[1e-5, 2e-5], [2e-5, 3e-5], [3e-5, 4e-5]]), # shape = (3,2)
)

outgoing = quadrupole.track(incoming)
```
produces an outgoing beam with `outgoing.particles.shape == (3, 2, 100_000, 7)` where the beamsize and quadrupole length are changed along each batch dimension (3,2) and the quadrupole strength is changes along the first dimension.

Note: Current implementation requires the same tensor dimensions for specifying beam attributes using `from_parameters`.

## Motivation and Context
Addresses #138 

## Types of changes

## Checklist


roussel-ryan:
If I can get feedback on the changes made here I can go ahead and make changes for other elements

jank324:
> If I can get feedback on the changes made here I can go ahead and make changes for other elements

So, I managed to break it. I really don't know what will be the best way to implement this automatic broadcasting. Ideally, we would find a module in PyTorch itself that does something similar and copy what it does.

Until then, maybe we should follow a test-driven approach, and implement tests that use Cheetah the way we would like to use it and then fix them.

roussel-ryan:
This is the current form of the dipole test that passes in my version of the repo. I can correct the rest of the tests if this looks ok to you all. @jank324 does this give you a clearer idea of what we meant by batching? Also what are the thoughts on using pytest to check that errors are raised? Is there something else already being used?
```
import pytest
import torch

from cheetah import Dipole, Drift, ParameterBeam, ParticleBeam, Segment


def test_dipole_off():
    """
    Test that a dipole with angle=0 behaves still like a drift.
    """
    dipole = Dipole(length=torch.tensor(1.0), angle=torch.tensor(0.0))
    drift = Drift(length=torch.tensor(1.0))
    incoming_beam = ParameterBeam.from_parameters(
        sigma_xp=torch.tensor(2e-7), sigma_yp=torch.tensor(2e-7)
    )
    outbeam_dipole_off = dipole(incoming_beam)
    outbeam_drift = drift(incoming_beam)

    dipole.angle = torch.tensor(1.0, device=dipole.angle.device)
    outbeam_dipole_on = dipole(incoming_beam)

    assert dipole.name is not None
    assert torch.allclose(outbeam_dipole_off.sigma_x, outbeam_drift.sigma_x)
    assert not torch.allclose(outbeam_dipole_on.sigma_x, outbeam_drift.sigma_x)


def test_dipole_batched_execution():
    """
    Test that a dipole with batch dimensions behaves as expected.
    """
    incoming = ParticleBeam.from_parameters(
        num_particles=torch.tensor(100),
        energy=torch.tensor(1e9),
        mu_x=torch.tensor(1e-5),
    )

    # test batching to generate 3 beam lines
    segment = Segment([
        Dipole(
            length=torch.tensor([0.5, 0.5, 0.5]),
            angle=torch.tensor([0.1, 0.2, 0.1]),
        ),
        Drift(length=torch.tensor(0.5)),
    ]
    )
    outgoing = segment(incoming)

    assert outgoing.particles.shape == torch.Size([3, 100, 7])
    assert outgoing.mu_x.shape == torch.Size([3])

    # Check that dipole with same bend angle produce same output
    assert torch.allclose(outgoing.particles[0], outgoing.particles[2])

    # Check different angles do make a difference
    assert not torch.allclose(outgoing.particles[0], outgoing.particles[1])

    # test batching to generate 18 beamlines
    segment = Segment([
        Dipole(
            length=torch.tensor([0.5, 0.5, 0.5]).reshape(3, 1),
            angle=torch.tensor([0.1, 0.2, 0.1]).reshape(1, 3),
        ),
        Drift(length=torch.tensor([0.5, 1.0]).reshape(2, 1, 1)),
    ]
    )
    outgoing = segment(incoming)
    assert outgoing.particles.shape == torch.Size([2, 3, 3, 100, 7])

    # test improper batching -- this does not obey torch broadcasting rules
    segment = Segment([
        Dipole(
            length=torch.tensor([0.5, 0.5, 0.5]).reshape(3, 1),
            angle=torch.tensor([0.1, 0.2, 0.1]).reshape(1, 3),
        ),
        Drift(length=torch.tensor([0.5, 1.0]).reshape(2, 1)),
    ]
    )
    with pytest.raises(RuntimeError):
        segment(incoming)
```

jank324:
> Also what are the thoughts on using pytest to check that errors are raised? Is there something else already being used?

I would say, that's how you are supposed to do it? Or is there a reason you wouldn't?

I think something very similar is already done somewhere in the Cheetah tests, but I'm not sure where at this point.

roussel-ryan:
Ok, that's how I would do it. I wasn't sure if pytest was being used or not and I didn't want to add a dependency if you had something else you wanted to use

jank324:
So for my understanding

```python
# Gives me the "product" of all settings
Dipole(
    length=torch.tensor([0.5, 0.5, 0.5]).reshape(3, 1),
    angle=torch.tensor([0.1, 0.2, 0.1]).reshape(1, 3),
)
...
assert outgoing.particles.shape == (3, 3, num_particles, 7)  # passes

# Matches the settings one-to-one
Dipole(
    length=torch.tensor([0.5, 0.5, 0.5]),
    angle=torch.tensor([0.1, 0.2, 0.1]),
)
...
assert outgoing.particles.shape == (3, num_particles, 7)  # passes
```

Correct?

roussel-ryan:
Yes, that is correct

jank324:
> Yes, that is correct

Seems like a good way of doing it to me then. Are there any places this could cause problems? And is there a straight-forward way of implementing this relatively easily for every element? ... such that we can give as a general guideline when we are implementing new elements, you should do it like ... and then you can be sure this will follow those rules?

roussel-ryan:
The main requirement is that you need to use pytorch operations which are broadcastable. For example, most of the changes that I needed to make (so far) is replacing the boolean indexing that you have been using for avoiding div0 errors with `torch.where` operations.
For example,
```
    gamma = energy / electron_mass_eV.to(device=device, dtype=dtype)
    igamma2 = torch.where(gamma == 0.0, 0.0, 1 / gamma**2)
```

roussel-ryan:
Note that I added a utility function to calculate the relativistic factors that are used throughout different elements

roussel-ryan:
I think removing the broadcasting from the beams will raise issues, we can discuss tomorrow

cr-xu:
I think changing `any` to `torch.any` would already fix some of the tests

roussel-ryan:
@jank324 I was able to fix the remaining tests EXCEPT for the issues associated with generating a uniform 3d ellipsoid, I think you know more about the inner workings of this function (which is raising an error not associated with batching I think)

cr-xu:
I fixed the issue but as before I couldn't push to the forked PR, so I pushed to a new branch [fix-non-batched-error](https://github.com/desy-ml/cheetah/tree/fix-non-batched-error) I guess @jank324 needs to merge the changes

tests/test_vectorized.py
```diff
@@ -325,133 +322,8 @@ def test_enormous_through_ares():
...
-    broadcast_element = element.broadcast((3, 10))
-
-    assert broadcast_element.length.shape == (3, 10)
-    assert broadcast_element.k1.shape == (3, 10)
-    for i in range(3):
-        for j in range(10):
-            assert broadcast_element.length[i, j] == 0.4
-            assert broadcast_element.k1[i, j] == 4.2
+    assert outgoing.energy.shape == torch.Size([1])
+    assert outgoing.total_charge.shape == torch.Size([1])
```
jank324, line 326:
```suggestion
    assert outgoing.energy.shape == (3, 100_000)
    assert outgoing.total_charge.shape == (3, 100_000)
```
I think these should actually also come out in the broadcast shape. Image running a `Cavity` at different settings. You would get different beams with different energies. So every `Element` should give you different energies for different settings, even if they end up being the same value.

The reason the broadcast fails here is because most elements will just copy over the energy. So we probably need a broadcasting alternative to that or just do it by hand.

jank324:
@roussel-ryan could you try to untick and then re-tick the _"Maintainers are allowed to edit this pull request."_ option on the right? We are trying to figure out why @cr-xu  can't push to this PR even though he should be able to.

roussel-ryan:
try it now

cr-xu:
By the way, I still couldn't push. I have no idea what's the problem.

roussel-ryan:
@jank324 so I don't think this is the right approach for broadcasting the energy, if an element alters the reference energy in a broadcastable way, it should happen when it modifies the reference energy

jank324:
> @jank324 so I don't think this is the right approach for broadcasting the energy, if an element alters the reference energy in a broadcastable way, broadcasting should happen when it modifies the reference energy

Hmm ... so the thing is: Right now only `Cavity` has an effect on the energy. So here the broadcasting I think just works.

The other elements so far just copy it over to the new beam. This leads to the awkward situation, where some properties of the new beam have a different vectorisation shape than others, e.g. `particles` may have become `(3, 2, ...)`, while `energy` is still at `(3, ...)`.

The `expand_as` approach right now only covers the case where the original energy was only a single value. So it actually wouldn't work for the example I just described. I'm just not quite sure what would ... other than doing weird things like this
```python
energy = energy + 0 * element.length
```

roussel-ryan:
what is the transform_to method supposed to do? Its not used anywhere in the package

jank324:
> [:, :, :

This allows you to take an existing particle distribution and change its properties. We originally implemented it to help with the domain randomisation in RL, where we had an incoming particle distribution coming from ASTRA and wanted to change its size, position etc.

roussel-ryan:
ok, I can fix the test then if its a useful feature. Also, I think supporting broadcasting between input parameters to these distribution creation methods is out of scope for this PR. Ie, if one wants to specify a batch of parameters for one parameter they must make sure the rest of the parameters are the same size. We can handle broadcasting in the future but I think we need to make more drastic changes to the beam creation methods to make that happen --> should go in a new PR IMO

roussel-ryan:
@jank324 All of the tests are passing now, I'm of the opinion that we should get this buttoned up so that we can start using it in GPSR and fix issues as they arise

cr-xu (commented):
There are some remaining tests that still use the broadcast, but apart from that, it looks ready and I agree that we should merge it and find issues afterwards.

```
FAILED tests/test_split.py::test_drift_end - AttributeError: 'ParticleBeam' object has no attribute 'broadcast'
FAILED tests/test_split.py::test_quadrupole_end - AttributeError: 'ParticleBeam' object has no attribute 'broadcast'
FAILED tests/test_split.py::test_cavity_end - AttributeError: 'ParticleBeam' object has no attribute 'broadcast'
FAILED tests/test_split.py::test_solenoid_end - AttributeError: 'ParticleBeam' object has no attribute 'broadcast'
FAILED tests/test_split.py::test_dipole_end - AttributeError: 'ParticleBeam' object has no attribute 'broadcast'
FAILED tests/test_split.py::test_undulator_end - AttributeError: 'ParticleBeam' object has no attribute 'broadcast'
FAILED tests/test_vectorized.py::test_screen_length_broadcast_shape - AttributeError: 'Screen' object has no attribute 'broadcast'
FAILED tests/test_vectorized.py::test_vectorized_undulator - AttributeError: 'ParticleBeam' object has no attribute 'broadcast'
FAILED tests/test_vectorized.py::test_vectorized_solenoid - AttributeError: 'ParticleBeam' object has no attribute 'broadcast'
```

roussel-ryan:
should be fixed

roussel-ryan:
Tests and functionality have been updated, a few notes on this commit:
- removes misalignment functionality in screen object (should be re-introduced in a future PR)
- allows beam to pass through screen for downstream tracking (useful functionality for beamline design, preserves consistency with other beam dynamics codes)
- raises NotImplemented Error for batching calculations using normal histogramming (couldn't figure out an efficient batching method that works for arbitrary batch dims)
- refactors from_astra particle imports to remove unnecessary leading batch dimension
Note: we should not expand the scope of this PR with too many additional features that can be introduced later

For some reason the following code introduces an unnecessary batch dimension into the lattice, I'm not sure where its coming from:
```
    cell = cheetah.converters.ocelot.subcell_of_ocelot(
        ares.cell, "AREASOLA1", "AREABSCR1"
    )
    segment = cheetah.Segment.from_ocelot(cell)
```
@jank324 can you take a look?

jank324:
> Tests and functionality have been updated, a few notes on this commit:
> 
> * removes misalignment functionality in screen object (should be re-introduced in a future PR)
> * allows beam to pass through screen for downstream tracking (useful functionality for beamline design, preserves consistency with other beam dynamics codes)
> * raises NotImplemented Error for batching calculations using normal histogramming (couldn't figure out an efficient batching method that works for arbitrary batch dims)
> * refactors from_astra particle imports to remove unnecessary leading batch dimension
>   Note: we should not expand the scope of this PR with too many additional features that can be introduced later
> 
> For some reason the following code introduces an unnecessary batch dimension into the lattice, I'm not sure where its coming from:
> 
> ```python
> cell = cheetah.converters.ocelot.subcell_of_ocelot(
>     ares.cell, "AREASOLA1", "AREABSCR1"
> )
> segment = cheetah.Segment.from_ocelot(cell)
> ```
> 
> @jank324 can you take a look?

I will take a look this coming week. No misalignments in `Screen` will be problematic though. They are a critical requirement for most of my RL stuff. I also don't like temporarily deprecated features, though in a `0.x` version this might be okay in really bad cases.

roussel-ryan:
Ok, I'll take a closer look at the misalignment stuff then. We can also completely remove basic histogramming if you don't want to raise the NotImplementedError, or we can also do the histogramming in serial

roussel-ryan:
I fixed the misalignment code, but batching with torch hist is still not possible atm since there is some missing functionality in pytorch. We could potentially modify the kde code to use a different kernel to approximate the behavior but that can come in a future PR IMO.

roussel-ryan:
@jank324 do you want me to fix the tests?

jank324:
> @jank324 do you want me to fix the tests?

No need. I already have them mostly fixed locally. Just haven't committed and pushed it yet.

cheetah/accelerator/element.py
```diff
@@ -91,9 +91,16 @@ def forward(self, incoming: Beam) -> Beam:
...
+    @property
+    def batch_shape(self) -> torch.Size:
+        tensors = []
+        # Get all parameters
+        for param in self.parameters():
+            tensors.append(param)
+        # Get all buffers
+        for buffer in self.buffers():
+            tensors.append(buffer)
+        return torch.broadcast_tensors(*tensors)[0].shape
```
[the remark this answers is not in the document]
cr-xu: Probably individual sub-classes should overwrite it if they have inherent non-scalar parameters.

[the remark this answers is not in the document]
jank324: I also stumbled over this and I am reviewing it at the moment. I only saw it used in a test once. I think overwriting it would be too much effort. I'm inclined to just remove it.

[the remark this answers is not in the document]
roussel-ryan: I think we can remove, and add in as necessary, it would be a good utility to have if possible

jank324:
What shape would we expect `outgoing.energy` to be? The original shape of `incoming.energy` or the broadcast shape of `outgoing.particles` ... i.e. `(3, 2, ...)`?

The former case is actually the kind of case that I was worried about, where it's possible that different properties of the result of a Cheetah beam have completely different vectorisation dimensions.

```python
@pytest.mark.parametrize(
    "ElementClass",
    [
        cheetah.Cavity,
        cheetah.Dipole,
        cheetah.Drift,
        cheetah.HorizontalCorrector,
        cheetah.Quadrupole,
        cheetah.RBend,
        cheetah.Solenoid,
        cheetah.TransverseDeflectingCavity,
        cheetah.Undulator,
        cheetah.VerticalCorrector,
    ],
)
def test_drift_broadcasting_two_different_inputs(ElementClass):
    """
    Test that broadcasting rules are correctly applied to a elements with two different
    input shapes for elements that have a `length` attribute.
    """
    incoming = cheetah.ParticleBeam.from_parameters(
        num_particles=100_000, energy=torch.tensor([154e6, 14e9])
    )
    element = ElementClass(length=torch.tensor([[0.6], [0.5], [0.4]]))

    outgoing = element.track(incoming)

    assert outgoing.particles.shape == (3, 2, 100_000, 7)
    assert outgoing.particle_charges.shape == (100_000,)
    assert outgoing.energy.shape == (3, 2)
```

roussel-ryan:
I think even in this edge case it's ok if the outgoing beam energy has a batch shape of just (3,) while the particle batch shape is (3,2). It understandable that the beam energy is not dependent on the drift length, while it would be dependent on batching in an rf cavity for example.

cheetah/accelerator/space_charge_kick.py
```diff
@@ -506,9 +505,9 @@ def _compute_forces(
         # Keep dimensions, and set F to zero if non-valid
         force_indices = (
             idx_vector,
-            torch.clamp(idx_x, max=grid_shape[0] - 1),
-            torch.clamp(idx_y, max=grid_shape[1] - 1),
-            torch.clamp(idx_tau, max=grid_shape[2] - 1),
+            torch.clamp(idx_x, min=0, max=grid_shape[0] - 1),
+            torch.clamp(idx_y, min=0, max=grid_shape[1] - 1),
+            torch.clamp(idx_tau, min=0, max=grid_shape[2] - 1),
         )
```
jank324, line 511: @RemiLehe @cr-xu I got negative indices here, which caused issues with a beam from ARES that I imported from Astra. Is the clamping "legal" or am I masking a different issue by doing that.

In addition, @cr-xu, you added that comment. I'm not sure I fully understand it. Is it correct / does it match what the code does?

cr-xu: I don't immediately see the issue here, if the test passes then it should be fine... 
Anyway I'm surprised that negative indices even appear. 

The comment refers to L507-518, basically masking the forces so that they come from only real particles. I 'm not sure if we need to move it.

tests/test_speed_optimizations.py
```diff
@@ -49,18 +49,18 @@ def test_merged_transfer_maps_tracking_vectorized():
     """
     incoming_beam = cheetah.ParameterBeam.from_astra(
         "tests/resources/ACHIP_EA1_2021.1351.001"
-    ).broadcast((10,))
+    )
```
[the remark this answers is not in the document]
jank324: So when Ryan updated this to the new broadcasting semantics, he seems to have removed the old vectorisation without properly adding a new one. I added a new vectorisation following the new semantics now the way I think it should be.

jank324 (approved):

jank324:
I fixed all the issues I could just see looking at the code changes and did a bit of cleanup on the way. It turned out to be quite a bit more than I expected.

Either way, I think as it is now, we can merge the PR.

cheetah/accelerator/transverse_deflecting_cavity.py
```diff
@@ -170,16 +170,24 @@ def _track_bmadx(self, incoming: ParticleBeam) -> ParticleBeam:
             )
         )
 
-        px = px + voltage * torch.sin(phase)
+        # TODO: Assigning px to px is really bad practice and should be separated into
+        # two separate variables
+        px = px + voltage.unsqueeze(-1) * torch.sin(phase)
```
cr-xu, line 175: Why do we need `.unsqueeze(-1)` here and (I think also the quadrupole bmad tracking)? I'm probably just confused.

W.r.t. the variable naming, it can probably be `px_new` or `px_out`

jank324: Because `px` has a shape`(vector, num_particles)` and `voltage` has `(vector, )` ... for PyTorch broadcasting the vector dimension of `voltage` basically has to be pushed to the left ... and that's what the `unsqueeze` does.

All the Bmad-X code needs refactoring, including that variable name ... I've had to keep myself from doing that in this PR.

cr-xu: I see! Then we should merge this first and do the refactoring later.

tests/test_kde.py
```diff
@@ -56,8 +58,50 @@ def test_weighted_samples_2d():
...
+def test_kde_1d_enforce_bins_shape():
+    """
+    Test that the 1D KDE histogram implementation correctly enforces the shape of the
+    bins tensor, i.e. throws an error if the bins tensor has the wrong shape.
+    """
+    data = torch.randn(100)  # 5 beamline states, 100 particles in 1D
+    bins = torch.linspace(0, 1, 10)  # A single histogram
+
+    with pytest.raises(ValueError):
+        kde_histogram_1d(data, bins, torch.rand(3) + 0.1)
```
cr-xu, line 89: I'm a bit confused about what the "5 beamline states" comment refers to.

jank324: It seems the vector dimension was intended but missing. Now it makes sense, right?

PR #209, opened 2024-07-04
Implement moving to `device` and changing `dtype`

### LINKED ISSUE #113: Moving elements and beams to devices doesn't work

jank324:
## Description

Implements that elements and beams can easily be moved to different devices and their beam types can be changes, like you would expect from a normal `torch.nn.Module`.

## Motivation and Context

This is normal for `torch.nn.Module` and would make working with different devices and dtypes in Cheetah much easier. Closes #113.

## Types of changes

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - Add Python 3.12 support (see #161) (@jank324)
 - Implement space charge using Green's function in a `SpaceChargeKick` element (see #142) (@greglenerd, @RemiLehe, @ax3l, @cr-xu, @jank324)
 - `Segment`s can now be imported from Bmad to devices other than `torch.device("cpu")` and dtypes other than `torch.float32` (see #196, #206) (@jank324)
+- Moving `Element`s and `Beam`s to a different `device` and changing their `dtype` like with any `torch.nn.Module` is now possible (see #209) (@jank324)
 
 ### 🐛 Bug fixes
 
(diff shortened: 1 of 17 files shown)


jank324:
@cr-xu @ansantam if we do something like this in an Element's `__init__`

```python
self.register_buffer(
    "k1",
    (
        torch.as_tensor(k1, **factory_kwargs)
        if k1 is not None
        else torch.zeros_like(self.length)
    ),
)
```

everything just works. It almost like the PyTorch people thought of this. 😄💡

jank324:
Can lines like
```python
factory_kwargs = {"device": device, "dtype": dtype}
```
then be removed?

jank324:
Should `Aperture.is_active`, `Aperture.shape` and `Aperture.lost_particles` be registered as buffers?

jank324:
> Can lines like
> 
> ```python
> factory_kwargs = {"device": device, "dtype": dtype}
> ```
> 
> then be removed?

In `torch.nn.Linear`, `factory_kwargs` are also used (in fact that's where I got it from). So I guess it should be considered PyTorch best practice.

jank324:
@cr-xu please review carefully. This seemed too easy.

cr-xu (approved):
Looks good to me. Since all the tests are passing let's go with it. I guess we'll see in practice if anything breaks.

PR #213, opened 2024-07-09
Fix twiss plot

jank324:
## Description

Fixes a bug in the Twiss plot.

## Motivation and Context

The Twiss plot was failing because it was not yet made compatible with vectorised Cheetah.

## Types of changes

## Checklist


cheetah/accelerator/segment.py
```diff
@@ -511,10 +511,10 @@ def plot_twiss(self, beam: Beam, ax: Optional[Any] = None) -> None:
             outgoing = element.track(longitudinal_beams[-1])
```
cr-xu, line 511: So now the `plot_twiss` only works for segment with shape (1,), right? Probably should add a warning or explicitly raise some error if calling with batched segment

jank324: Yeah ... actually right now **all** plot functions only use the first beam in a batch. I'm not really a fan of cluttering things with warnings (just because I know what that will look like). So either we at least mention it in the docs, and maybe add a warning, but with an option to turn that warning off?

jank324: Or we add an argument to all plot function where you give the index you want from the batch.

cr-xu: Yes, we should at least mention it in the doc, and have a default behavior when passing a batched segment, because now this will just fail.

cr-xu: > Or we add an argument to all plot function where you give the index you want from the batch.

Or... I just realised we don't have a convenient function to return one instance of the batched segment yet

jank324: > > Or we add an argument to all plot function where you give the index you want from the batch.
> 
> Or... I just realised we don't have a convenient function to return one instance of the batched segment yet

Like an `unbroadcast`? 😄

jank324: I'm also not sure how all of this would play with the broadcast rules Ryan proposed.

cr-xu: > > > Or we add an argument to all plot function where you give the index you want from the batch.
> > 
> > 
> > Or... I just realised we don't have a convenient function to return one instance of the batched segment yet
> 
> Like an `unbroadcast`? 😄

Sure, something like a `__getitem__(index)`. The potential usecases would be 
1. Plotting 
2. Pass a batch of settings, get a batch of output beam, and retrieve the best setitng with `segment[torch.argmax(some_beam_parameter)]`

jank324: Yeah ... we should have this. I worry we should wait until the broadcasting is in to do that though.

Hespe:
Plotting currently still fails if the segment is broadcasted beforehand. The problem is caused by the incompatible splitting implementation (#227).

tests/test_plotting.py
```diff
@@ -57,3 +57,53 @@ def test_reference_particle_plot():
...
+    segment.AREAMCVM1.k1 = torch.tensor(1e-3)
+    segment.AREAMQZM3.k1 = torch.tensor(5.0)
+    segment.AREAMCHM1.k1 = torch.tensor(-2e-3)
+
+    incoming = cheetah.ParticleBeam.from_astra(
+        "tests/resources/ACHIP_EA1_2021.1351.001"
+    )
+
+    # Run the plotting to see if it raises an exception
+    segment.plot_twiss(incoming, vector_idx=(0, 2))
```
cr-xu, line 84: That's still in progress or @jank324  you forgot to push the updates?

jank324: I'm working on it right now.

cr-xu (approved):

PR #215, opened 2024-07-09
Fix `Cavity` being broken by vectorisation

### LINKED ISSUE #214: Vectorised cavity is broken
Tracking through the following cavity results in NaNs (in the transfer map `tm`)

```python
cavity = cheetah.Cavity(
    length=torch.tensor([3.0441]),
    voltage=torch.tensor([0.0]),
    phase=torch.tensor([-0.0]),
    frequency=torch.tensor([2.8560e09]),
    name="k27_1a",
    dtype=torch.float64,
)
```

This is caused by the voltage being `0.0`.

It seems like we suspected that because there is this comment in the `Cavity` code

```python
def transfer_map(self, energy: torch.Tensor) -> torch.Tensor:
    # There used to be a check for voltage > 0 here, where the cavity transfer map
    # was only computed for the elements with voltage > 0 and a basermatrix was
    # used otherwise. This was removed because it was causing issues with the
    # vectorisation, but I am not sure it is okay to remove.
    tm = self._cavity_rmatrix(energy)

    return tm
```

I suspect it wasn't okay to remove.

jank324:
## Description

Re-adds the use of `base_rmatrix` for `Cavity`s with `voltage=0.0` because the usually cavity matrix breaks in that case. This was originally removed with the vectorisation because it was difficult to integrate and though to be unnecessary.

## Motivation and Context

Closes #214.

## Types of changes

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### 🚨 Breaking Changes
 
-- Cheetah is now vectorised. This means that you can run multiple simulations in parallel by passing a batch of beams and settings, resulting a number of interfaces being changed. For Cheetah developers this means that you now have to account for an arbitrary-dimensional tensor of most of the properties of you element, rather than a single value, vector or whatever else a property was before. (see #116, #157, #170, #172, #173, #198) (@jank324, @cr-xu)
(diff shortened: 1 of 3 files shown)


cr-xu:
The failing test is actually related to the broadcast defaulting to float32 #217

jank324:
> The failing test is actually related to the broadcast defaulting to float32 #217

I've merged #218 and now they are passing.

cr-xu (approved):

Issue #221, opened 2024-07-18
Add elegant lattice compatibility


Hespe:
In addition to the currently existing lattice converters, it would be useful to have an implementation for importing from elegant lattices.
Examples files are available [here](https://www.aps.anl.gov/Accelerator-Operations-Physics/Software).

PR #222, opened 2024-07-18
Add Elegant Lattice Converter

Hespe:
## Description

Add an import function for lattices in the elegant file format. Since they are closely related to Bmad lattices, many of the existing functions can be reused.

## Motivation and Context

The converter is not handling all elements defined in elegant, but is handles all that are required to import an EuXFEL electron dump lattice. For some elements there is so far no perfect match so far:

ToDos:
- [x] ~~Sextupole magnet implementation~~ (#42, #230)
- [ ] Validate custom transfer map (elegant `ematrix`)
- [ ] Decide about additional cavity parameters
- [ ] Check if solenoid and kicker are assigned to the correct element types
- [x] Handle `SBend`s with `k1` property (#235)

- [x] I have raised an issue to propose this change (#221)

## Types of changes

## Checklist


Hespe:
Right now, I simply imported the relevant functions from the Bmad converter into the elegant file. It might be more appropriate to extract them into a third file, since they are as closely related to elegant as they are to Bmad.

Hespe:
Some sector bends one encounters in elegant lattice files have both a steering angle and a focussing strength. Such an element is currently missing in Cheetah, see #235.

Hespe:
According to the description, there are still four ToDos left. The `Sextupole` parsing can be corrected later once an implementation is ready, I guess, leaving only validation tasks.

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 - Moving `Element`s and `Beam`s to a different `device` and changing their `dtype` like with any `torch.nn.Module` is now possible (see #209) (@jank324)
 - `Quadrupole` now supports tracking with Cheetah's matrix-based method or with Bmad's more accurate method (see #153) (@jp-ga, @jank324)
 - `Dipole` now takes a focusing moment `k1` (see #235) (@hespe)
+- Implement a converter for lattice files imported from Elegant (see @222) (@hespe)
```
jank324, line 21:
```suggestion
- Implement a converter for lattice files imported from Elegant (see #222) (@hespe)
```

tests/test_elegant_conversion.py
```diff
@@ -0,0 +1,95 @@
...
+        name="fodo",
+    )
+
+    assert converted.name == correct_lattice.name
+    assert [element.name for element in converted.elements] == [
+        element.name for element in correct_lattice.elements
+    ]
+    assert converted.q1.length == correct_lattice.q1.length
+    assert converted.q1.k1 == correct_lattice.q1.k1
+    assert converted.q2.length == correct_lattice.q2.length
```
jank324, line 35: This test is missing a docstring. 

I'm also a little surprised that it works despite properties only being defined as scalars. I think they should be defined as single-element vectors for now (even though #208 should hopefully and eventually remove the need for that).

jank324 (changes requested):
I think once the test is fixed, we are fine.

tests/test_elegant_conversion.py
Hespe: I think this works since the elements properties are only read when calling `segment.track()`, and constructing scalar valued elements is perfectly fine. But of course you are right that the test should be done properly.

jank324 (approved):

jank324:
Ah ... I named the shared code file for the Bmad and Elegant converters `fortran_namelist` now. I'm not perfectly sure if this is correct. It could also be `sdds` ... maybe ... I guess. If someone knows better, let me know.

cr-xu:
> Ah ... I named the shared code file for the Bmad and Elegant converters `fortran_namelist` now. I'm not perfectly sure if this is correct. It could also be `sdds` ... maybe ... I guess. If someone knows better, let me know.

I think they are not the standard Fortran namelists (compared to ASTRA which is just fortran 90)

I would suggest just calling the file something like `elegant_bmad_namelists`

Btw: bmad has an "from elegant" converter implemented in python, maybe that's a good point to check:
- https://github.com/bmad-sim/bmad-ecosystem/blob/main/util_programs/elegant_to_bmad/elegant_to_bmad.py

cr-xu (approved):

Issue #184, opened 2024-06-18
Add `flake8_numbers`


jank324:
We should probably start using https://github.com/frequenz-floss/flake8-numbers

jank324:
Closing this for the same reason the corresponding PR #224 was closed.

PR #224, opened 2024-07-19
Add flake8-numbers plugin to formatting step

Hespe:
## Description

This change adds the flake8-numbers plugin to ensure all number literals are properly formatted.
The `tests/resources` and `docs/examples` directories are excluded, since they contain lattice files that should not be linted.

## Motivation and Context

- [x] I have raised an issue to propose this change (#184)

## Types of changes

## Checklist


.github/workflows/format.yaml
```diff
@@ -11,7 +11,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: "3.10"
+          python-version: "3.11"
```
jank324, line 14:
```suggestion
          python-version: "3.12"
```

jank324:
It seems a little like `flake8-numbers` doesn't work all that well yet.

Technically, Python doesn't care where you put digit separators. It seems `flake8-numbers` does and it wants you to do e.g. `3.14_159_265_359` But I feel like `3.141_592_653_59` is more intuitive(?) The latter seems to also be what GitHub Copilot wants to do.

Hespe:
I agree, I would have preferred to start counting the digits from the decimal separator, to the front and to the back. Starting from the last digit seems wrong.

jank324:
I converted this back to draft for now, because with the current state of the `flake8-numbers` plugin, I think we should defer on using it.

jank324:
Closing this because the plug-in is not in a state for us to use it.

PR #240, opened 2024-08-20
Add Bmad-X Drift Dipole and Transverse Deflecting Cavity

jp-ga:
## Description

Adds Bmad-X tracking routines to `Drift` and `Dipole`, and adds a new `Element`: a `TransverseDeflectingCavity`

## Motivation and Context

## Types of changes

## Checklist

- [ x] I have updated the tests accordingly (_required for a bug fix or a new feature_).
  
  
- [ x] I have reformatted the code and checked that formatting passes (**required**).
- [ x] I have have fixed all issues found by `flake8` (**required**).
- [ x] I have ensured that all `pytest` tests pass (**required**).


cheetah/accelerator/dipole.py
```diff
@@ -1,45 +1,62 @@
...
+    :param gap: The magnet gap [m], NOTE in MAD and ELEGANT: HGAP = gap/2.
+    :param gap_exit: The magnet gap at the entrance [m],
+        NOTE in MAD and ELEGANT: HGAP = gap/2.
     :param name: Unique identifier of the element.
     """
 
     def __init__(
         self,
         length: Union[torch.Tensor, nn.Parameter],
+        p0c: Optional[Union[torch.Tensor, nn.Parameter]] = None,
```
cr-xu, line 46: I think the dipole should also get the reference momentum from particles with `ParticleBeam.p0c`  rather than defining it as its own parameter. Is there a specific use case that you want to have dipole ref momentum other than the particle one?
https://github.com/desy-ml/cheetah/blob/c909024359dee8d2479d4ca4534d743ec974c1ab/cheetah/particles/beam.py#L299

cheetah/accelerator/dipole.py
```diff
@@ -1,47 +1,69 @@
...
-    :param gap: The magnet gap [m], NOTE in MAD and ELEGANT: HGAP = gap/2
+    :param fringe_integral_exit: Fringe field integral of the exit face. Only set if
+        different from `fringe_integral`.
+    :param fringe_at: Where to apply the fringe fields. The available options are:
+        - "both_ends": Apply fringe fields at both ends.
+        - "entrance_end": Apply fringe fields at the entrance end.
+        - "exit_end": Apply fringe fields at the exit end.
+        - "no_end": Do not apply fringe fields.
+    :param fringe_type: Type of fringe field. Currently only supports `"linear_edge"`.
     :param name: Unique identifier of the element.
```
jank324, line 46: @jp-ga Do any of these only apply for `bmadx` tracking? If so, I think this should be mentioned in the docstring.

cheetah/accelerator/dipole.py
```diff
@@ -126,12 +165,170 @@ def hx(self) -> torch.Tensor:
...
+        outgoing_beam = ParticleBeam(
+            torch.stack((x, px, y, py, tau, delta, torch.ones_like(x)), dim=-1),
+            ref_energy,
+            particle_charges=incoming.particle_charges,
+            device=incoming.particles.device,
+            dtype=incoming.particles.dtype,
+        )
+        return outgoing_beam
+
+    def _bmadx_body(self, x, px, y, py, z, pz, p0c_particle, mc2):
```
jank324, line 248: `_bmadx_body` is missing type annotations and a docstring

cheetah/accelerator/dipole.py
```diff
@@ -126,12 +165,170 @@ def hx(self) -> torch.Tensor:
...
+        beta0 = self.p0c / E0
+
+        x_f = x2
+        px_f = px_norm * torch.sin(self.angle + phi1 - theta_p)
+        y_f = y + py * Lp / px_norm
+        z_f = z + (beta * self.length / beta0) - ((1 + pz) * Lp / px_norm)
+
+        return x_f, px_f, y_f, py, z_f, pz
+
+    def _bmadx_fringe_linear(
```
jank324, line 301: `_bmadx_fringe_linear` is missing a docstring and a return type annotation

cheetah/accelerator/transverse_deflecting_cavity.py
```diff
@@ -0,0 +1,251 @@
...
+        return [self]
+
+    def plot(self, ax: plt.Axes, s: float) -> None:
+        alpha = 1 if self.is_active else 0.2
+        height = 0.4
+
+        patch = Rectangle(
+            (s, 0), self.length[0], height, color="gold", alpha=alpha, zorder=2
+        )
+        ax.add_patch(patch)
```
jank324, line 229: Is it bad that `Cavity` and `TransverseDeflectingCavity` currently plot the same?

jank324 (changes requested):
So as far as I'm concerned, this PR is fine, except for the docstring and type annotation that's missing. @jp-ga, I think you are in the best position to add those.

I'm not sure if the `plot` function is an issue. @jp-ga, @cr-xu what are your thoughts?

cheetah/accelerator/transverse_deflecting_cavity.py
cr-xu: I would suggest to use a different color.

jp-ga:
> So as far as I'm concerned, this PR is fine, except for the docstring and type annotation that's missing. @jp-ga, I think you are in the best position to add those.
> 
> I'm not sure if the `plot` function is an issue. @jp-ga, @cr-xu what are your thoughts?
> 
> Aside from that, I fixed an issue that always cast everything to `torch.float64`. This was rooted in a general issue in Cheetah, where constant floating point numbers were cast to tensors when they didn't have to be.

Sounds good, I will work on the changes. I can also change the color of the TDC.

cheetah/accelerator/dipole.py
jp-ga: I think you're right. I don't see any useful cases in which the reference momentum of particles and elements should be different. I will make use of `ParticleBeam.p0c`

jank324 (approved):

Issue #248, opened 2024-09-04
cheetah.Aperture class does not include the length parameter


zihan-zh:
I found element length is not defined in the `cheetah.Aperture` object when converting the `collimator` element from Bmad lattice file.

PR #249, opened 2024-09-04
Fix Bmad collimator length on import

jank324:
## Description

Changes in interpretation of Bmad `ecollimator` and `rcollimator` in Cheetah from a lengthless `Aperture` element to a `Drift` + `Aperture` sequence.

## Motivation and Context

This was a bug found in the LCLS import that led to a Twiss mismatch. See #248.

## Types of changes

## Checklist


cr-xu:
I have only one minor concern: I guess when you have a real aperture with non-length, the beam should be smaller than some threshold over the whole length.
Would it make more sense to convert it to `[Aperture_start, Drift, Aperture_end]`?

cr-xu (approved):

jank324:
> I have only one minor concern: I guess when you have a real aperture with non-length, the beam should be smaller than some threshold over the whole length. Would it make more sense to convert it to `[Aperture_start, Drift, Aperture_end]`?
> 
> Edit: looking at the Bmad manual it happens only at the "exit end" by default
> 
> image

I was thinking about the same thing. In the end, only an infinite number of apertures would be truly correct. But this quick fix worked well enough to fix the mismatch I had. We can think about adding more options or even a `Collimator` element at a later date if this is not enough.

Issue #253, opened 2024-09-17
Fall back to PyTorch default dtype if no explicit type is provided


Hespe:
Currently, a number of methods use `torch.float32` as their default `dtype` if no explicit choice is provided. While this is often a reasonable choice, some applications require more (or less) precision, which the current default inhibits. Instead, the methods should query `torch.get_default_dtype()` such that the user can globally configure the desired level of precision.

Issue #238, opened 2024-08-19
Misleading docstring and type hints in beam class methods


Hespe:
The docstrings, parameter names and type hints of some `beam` class methods are not consistent with the new coordinate system. In particular, the `from_twiss()` method refers to `sigma_s` and its docstrings for `sigma_p` disagrees with `from_parameters()` about its unit.

Furthermore, the `num_particles` parameter of `ParticleBeam` methods is supposed to be a `torch.Tensor` according to its type hint, but the methods crash if a tensor is provided and a standard integer works.

Issue #165, opened 2024-05-28
Release `v0.7`


jank324:
# Checklist for `v0.7` release features

 - [x] Vectorisation (#116)
 - [x] Add space charge (#142)
 - [ ] ~~Add nice interface for space charge~~ (delayed to later release) (#160)
 - [x] Add custom transfer map element (#100)
 - [x] Fix issue with default length (#116, #143, #164)
 - [x] Add uniformly distributed beam initialisation (#146)
 - [x] Fix NumPy 2 compatibility (#177)

Let's discuss what features we want to have ready before releasing `v0.7`.

jank324:
One thought, now that we are a couple more people contributing to Cheetah, we should use this release to establish common ground on certain foundational design choices, e.g. the coordinate system. So if there is something like that you thing needs to be discussed, please bring it up here.

cr-xu:
From the discussion before, the current plan is to  

- Fix remaining issues, review the code and merge #142
- Adapt the changes and make sure SC implementation is still correct in #163, then merge it into master
- Implement #160

jank324:
I have some proposals on revising the plan for `v0.7.0`, considering that the version of Cheetah on `master` is by now very different from the most recent stable release you would get from PyPI ( `v0.6.3` without vectorisation and such).

 - #142 is merged.
 - #163 should definitely be merged before, being a bit of a breaking change and almost done.
 - I would suggest to defer #160 to `v0.7.1`
 - I think #208 I think is an important item we are working on right now, so it would be nice to have this in `v0.7.0`. But, considering that it will probably take some time, it might also make sense to make that part of `v0.7.1`.
 - #153 is almost ready, but it really doesn't matter if this appears in `v0.7.0` or `v0.7.1`.
 - I'm not sure where we stand with #164.
 - #213 is a bug fix for the vectorisation, so it should be merged before `v0.7.0` ... but it's ready, so that should be easy.

So I guess we could release `v0.7.0` as soon as #213 is merged and #163 is finalised and merged. The only open question is if we would want to wait for #208.

cr-xu:
I believe #164 is already fixed with #209.

jank324:
#227 will have to be fixed before release as well.

jank324:
#241 also needs fixing before this release!

jank324:
#246 also goes on the list.

cr-xu:
[...]

PR #254, opened 2024-09-17
Improve default dtype selection

Hespe:
## Description

Many methods currently default to creating tensors with `dtype=torch.float32` regardless of the data provided to them or the default `dtype` configured PyTorch. This PR changes those methods to either keep the same `dtype` as their arguments or fall back to `torch.get_default_dtype()` if no reasonable choice is available.
Similar changes are implemented for the `device` of said tensors.

Since this PR is changing the default arguments of a number of methods, it should be considered a _breaking change_. The impact is likely negligable because the new fallback `torch.get_default_dtype()` aligns with the previous default of `torch.float32` if no explicit actions are taken by the user.

## Motivation and Context

Currently, it is rather cumbersome to track particles with double precision since many methods default to `torch.float32`. Implementing this change will increase the compatability applications that require tracking using `torch.float64` (or `torch.float16`).

- [x] I have raised an issue to propose this change #253 

## Types of changes

## Checklist


Hespe:
I've noticed that the docstrings for some of the methods I have edited are not consistent with the method signatures, especially in case of the `ParticleBeam`. This is in addition to the issues already noted in #238

Hespe:
In principle, the methods for manual initialization of `Beam` or `Element` objects should be done. We probably also want to change the methods for importing lattices and beams defined in other tools, but I would argue its probably best to leave that for a separate PR since this one is already touching enough files as is.

For many of the methods, I currently only applied the minimal change of removing the explicit default of `torch.float32`, leaving the `dtype` selection to the built-in PyTorch functions. There might however be an issue with that approach. By supplying `dtype=None`, the default behaviour of the built-in functions is to keep the `dtype` of its input argument. Therefore, the `Element` might have conflicting `dtype` for its parameters if different types are passed to the constructors.
We could either say that it's the responsibility of the user to ensure a consistent `dtype` is used, or we have to try to infer the dtype on our own from all not `None` arguments, similar to the `from_twiss()` function for `Beam`.

Hespe:
@jank324 It seems to me like the remaining test failures are uncovered but not caused by the changes of this PR. The following code also crashes on `master`:

```python
incoming = cheetah.ParticleBeam.from_parameters(
    num_particles=100_000, energy=torch.tensor([154e6, 14e9])
)
element = cheetah.TransverseDeflectingCavity(
    length=torch.tensor(1.0),
    phase=torch.tensor([[0.6], [0.5], [0.4]]),
)
outgoing = element.track(incoming)
```

Maybe this is also relevant for #165 ?

cheetah/accelerator/transverse_deflecting_cavity.py
```diff
@@ -55,23 +58,23 @@ def __init__(
...
         self.register_buffer(
             "frequency",
             (
                 torch.as_tensor(frequency, **factory_kwargs)
                 if frequency is not None
-                else torch.tensor(0.0, **factory_kwargs)
+                else torch.zeros_like(self.length)
             ),
         )
         self.register_buffer(
```
jank324, line 80: I think these lines are causing / exposing the issue. They should definitely be like
```python
torch.tensor(0.0, **factory_kwargs)
```
And then we should check if vectorising things like `frequency` works, because vectorising them by mistake seems to have caused the test failures.

Hespe: I agree. This will likely fix the tests for this PR, but its effectively just hiding the issue.

Hespe:
Created #270 for the issue with the deflecting cavity.

jank324:
> Created #270 for the issue with the deflecting cavity.

I just merged the updated `master` with the fix for #270 in here.

jank324:
Okay, just to write down some todos here:
 - [ ] Fix non-necesary `nn.Parameter` type annotation
 - [ ] Fix non-buffer beam properties
 - [ ] Fix `ParticleBeam.uniform3dellipsoid`

cheetah/accelerator/cavity.py
```diff
@@ -288,12 +293,17 @@ def _cavity_rmatrix(self, energy: torch.Tensor) -> torch.Tensor:
...
+        beta1 = torch.tensor(1.0, **factory_kwargs)
 
-        k = 2 * torch.pi * self.frequency / torch.tensor(constants.speed_of_light)
-        r55_cor = torch.tensor(0.0)
+        k = (
+            2
+            * torch.pi
+            * self.frequency
+            / torch.tensor(constants.speed_of_light, **factory_kwargs)
+        )
```
jank324, line 305:
```suggestion
        k = (
            2
            * torch.pi
            * self.frequency
            / constants.speed_of_light
        )
```

cheetah/accelerator/screen.py
```diff
@@ -203,7 +196,9 @@ def reading(self) -> torch.Tensor:
         read_beam = self.get_read_beam()
         if read_beam is Beam.empty or read_beam is None:
             image = torch.zeros(
-                (int(self.effective_resolution[1]), int(self.effective_resolution[0]))
+                (int(self.effective_resolution[1]), int(self.effective_resolution[0])),
+                device=self.misalignment.device,
+                dtype=self.misalignment.dtype,
```
jank324, line 201: I don't quite understand the change here.

cheetah/accelerator/space_charge_kick.py
```diff
@@ -589,7 +587,11 @@ def track(self, incoming: ParticleBeam) -> ParticleBeam:
                 ],
                 dim=-1,
             )
-            cell_size = 2 * grid_dimensions / torch.tensor(self.grid_shape)
+            cell_size = (
+                2
+                * grid_dimensions
+                / torch.tensor(self.grid_shape, **self.factory_kwargs)
+            )
```
jank324, line 594: I'm not quite sure this is right.

cheetah/accelerator/screen.py
```diff
@@ -118,8 +108,11 @@ def is_skippable(self) -> bool:
         return not self.is_active
 
     @property
-    def effective_resolution(self) -> torch.Tensor:
-        return self.resolution / self.binning
+    def effective_resolution(self) -> tuple[int, int]:
+        return (
+            self.resolution[0] / self.binning,
+            self.resolution[1] / self.binning,
```
Hespe, line 114:
```suggestion
            int(self.resolution[0] / self.binning),
            int(self.resolution[1] / self.binning),
```

jank324 (approved):

cheetah/accelerator/screen.py
```diff
@@ -118,8 +108,11 @@ def is_skippable(self) -> bool:
         return not self.is_active
 
     @property
-    def effective_resolution(self) -> torch.Tensor:
-        return self.resolution / self.binning
+    def effective_resolution(self) -> tuple[int, int]:
+        return (
+            int(self.resolution[0] / self.binning),
+            int(self.resolution[1] / self.binning),
```
Hespe, line 114:
```suggestion
            self.resolution[0] // self.binning,
            self.resolution[1] // self.binning,
```

cheetah/accelerator/screen.py
```diff
@@ -203,7 +196,9 @@ def reading(self) -> torch.Tensor:
         read_beam = self.get_read_beam()
         if read_beam is Beam.empty or read_beam is None:
             image = torch.zeros(
-                (int(self.effective_resolution[1]), int(self.effective_resolution[0]))
+                (int(self.effective_resolution[1]), int(self.effective_resolution[0])),
```
Hespe, line 199:
```suggestion
                (self.effective_resolution[1], self.effective_resolution[0]),
```

The `int()` is unnecessary if the property satisfies its type hint.

Issue #216, opened 2024-07-09
Add `s` position to elements


jank324:
Right now elements don't know where they are and a `Segment` assumes it starts at 0 m. That might not always be desirable.

PR #268, opened 2024-10-03
Fix `Aperture` vectorisation issue

### LINKED ISSUE #93: What is the purpose of Beam.empty?
In the `ParticleBeam`, there are some locations where it's checked whether `self is not Beam.empty`, e.g. https://github.com/desy-ml/cheetah/blob/b92502966f01797889697f1dcba46a2667328f8c/cheetah/particles.py#L1023

But this is never the case, since one cannot initialize the ParticleBeam without any argument. Also, the `Beam.empty` is simply a str, rather than an actual empty Beam object. This is very confusing:

```
>>> emptybeam = ParticleBeam(torch.empty(1,7), energy=torch.tensor([])). # The best one can do to initialize ParticleBeam without anything?
>>> emptybeam is Beam.empty
False
```

### LINKED ISSUE #241: Vectorised aperture tracking is broken

jank324:
## Description

Add `survival_probabilities` property to `ParticleBeam`. The particles blocked by the `Aperture` will be marked with `0` and surviving particles will have `1`. 

Right now `Aperture` still blocks differentiability, but can be addressed later using a sigmoid function for a smoothed blocking.

## Motivation and Context

Fixes #241. Also closes #93.

## Types of changes

## Checklist


jank324:
We kind of already talked about this a little bit, but I think this will basically require that the `survived_mask` is moved to be a permanent property of `ParticleBeam` ... and then all other non-aperture computations will have to be changed to respect the mask.

cr-xu:
> We kind of already talked about this a little bit, but I think this will basically require that the `survived_mask` is moved to be a permanent property of `ParticleBeam` ... and then all other non-aperture computations will have to be changed to respect the mask.
> 
> Correct? Am I missing something?

Yes, the properties of the beam should be calculated only w.r.t. ParticleBeam[survived_mask]. 

Then we need to think of how other elements handle the tracking.
Probably it's easier to still propagate the whole beam (for the vectorized implementation) even if a portion of the macroparticles are already lost.

I would add another property like `particle_lost_at` to keep track of where the particles are lost along the segment. But this requires probably keeping track of the `s` positions as #216 proposed.

jank324:
So, I think we should possibly approach this PR slightly differently. A mask and the way apertures currently work are not gradient-friendly, because it's all step functions. Maybe we should instead consider a survival probability for each function. And then consider beam properties as survival probability-weighted. `Apterture` would then apply a smooth mirrored Sigmoid-like function to the probabilities, which would give us nice gradients.

Another handy thing about this would be that we could leave the masked implementation of `Aperture` in place for now and use the mask as probabilities in other computations. We then wouldn't have to implement the smooth masking now and could do that in a later PR if we wanted, but the rest of the code would already be able to work with it.

cr-xu:
This sounds good for me. I guess we'll see how this scales with collective effects later...

@jank324  will you finish this PR?

jank324:
> This sounds good for me. I guess we'll see how this scales with collective effects later...
> 
> @jank324 will you finish this PR?

I actually think it will scale very nicely. Masking always flattens tensors, so you have to run a bunch of reshaping with Python operations. Multiplying with probabilities is one multiplication operation that can completely run concurrently and in machine code, so should be super fast.

I meant to finish it myself, but right now I'm not on top enough of my submission schedule to dedicate time to Cheetah development. So if you want to work on this, go right ahead.

jank324:
Some todo notes for myself:
 - [ ] `Beam.empty` can probably be removed as part of this.
 - [ ] I guess `Screen` needs to account for lost particles as well.

jank324:
So ... the last thing this PR is missing now is that the `Screen` reading should account for the survival probabilities. As part of this, it should also account for the particle charges, so
```python
histogram_weights = read_beam.suvival_probabilities * read_beam.particle_charges
```

Issue #272, opened 2024-10-07
Misleading BPM import in the Elegant converter


Hespe:
The Elegant converter currently does not properly map the Elegant `MONI` to the `cheetah.BPM`. Furthermore, the beam position should be measured in the middle for monitors with non-zero length.

PR #273, opened 2024-10-07
Fix parsing of BPM and CustomTransferMap from Elegant

Hespe:
## Description

The Elegant `MONI` element is currently not properly mapped to the Cheetah `BPM`. Furthermore, there is a slight issue with how `MONI` with non-zero length are handled.
Similarly the current parser for `CustomTransferMap` is faulty since the `dtype` of the resulting tensor cannot be properly inferred.

## Motivation and Context

- [x] I have raised an issue to propose this change
  - #272 
  - #274 

## Types of changes

This PR is a breaking change, since elegant `MONI` elements are now mapped to `BPM` instead of `Marker`, requiring changes to downstream code.

## Checklist


tests/test_elegant_conversion.py
```diff
@@ -32,14 +33,42 @@ def test_fodo():
...
-    assert converted.s1.e1 == correct_lattice.s1.e1
+    assert torch.isclose(converted.q1.length, correct_lattice.q1.length)
+    assert torch.isclose(converted.q1.k1, correct_lattice.q1.k1)
+    assert torch.isclose(converted.q2.length, correct_lattice.q2.length)
+    assert torch.isclose(converted.q2.k1, correct_lattice.q2.k1)
+    for i in range(2):
+        assert torch.isclose(converted.d1[i].length, correct_lattice.d1[i].length)
+    assert torch.isclose(converted.d2.length, correct_lattice.d2.length)
+    assert torch.isclose(converted.s1.length, correct_lattice.s1.length)
+    assert torch.isclose(converted.s1.e1, correct_lattice.s1.e1)
```
jank324, line 44: I feel like this should be no loop and use `torch.allclose`.

tests/test_elegant_conversion.py
```diff
@@ -32,14 +33,42 @@ def test_fodo():
...
+            [0.0, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0],
+            [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0],
+        ]
+    )
+
+    assert np.isclose(converted.c1.length, 0.7)
+    assert np.isclose(converted.c1.frequency, 1.2e9)
+    assert np.isclose(converted.c1.voltage, 16.175e6)
+    for i in range(2):
+        assert torch.allclose(converted.c1e[i]._transfer_map, correct_map)
```
jank324, line 68:
```suggestion
    correct_transfer_map = torch.tensor(
        [
            [1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0],
            [0.04, 1.0, 0.003, 0.0, 0.0, 0.0, -0.0027],
            [0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0],
            [0.003, 0.0, -0.04, 1.0, 0.0, 0.0, -0.15],
            [0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0],
            [0.0, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0],
            [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0],
        ]
    )

    assert np.isclose(converted.c1.length, 0.7)
    assert np.isclose(converted.c1.frequency, 1.2e9)
    assert np.isclose(converted.c1.voltage, 16.175e6)
    for i in range(2):
        assert torch.allclose(converted.c1e[i]._transfer_map, correct_transfer_map)
```

tests/test_elegant_conversion.py
```diff
@@ -32,14 +33,42 @@ def test_fodo():
...
+    assert torch.isclose(converted.q2.k1, correct_lattice.q2.k1)
+    for i in range(2):
+        assert torch.isclose(converted.d1[i].length, correct_lattice.d1[i].length)
+    assert torch.isclose(converted.d2.length, correct_lattice.d2.length)
+    assert torch.isclose(converted.s1.length, correct_lattice.s1.length)
+    assert torch.isclose(converted.s1.e1, correct_lattice.s1.e1)
+
+
+def test_cavity_import():
+    """Test importing an accelerating cavity defined in the Elegant file format."""
```
jank324, line 48: I feel like this is a weird instance of cavity import and the docstring (and maybe even the name) of the test should reflect and explain that.

jank324: On second thoughts, this is really two separate tests, one testing if the cavity was imported correctly (and without throwing an exception) and another doing the save for a custom transfer map, right?

tests/test_elegant_conversion.py
Hespe: I`m not sure how `torch.allclose` would work here, but maybe there is some trick I`m missing. The problem is that we have two elements with name `d1` and therefore two separate `length` tensors, not one multidimensional tensor.

tests/test_elegant_conversion.py
Hespe: I`ve decided to split the test into two separate parts. One for testing the regular cavity import, the second for the custom transfer map.

tests/test_elegant_conversion.py
jank324: Oh I see ... I misunderstood the that line 🙃

jank324 (approved):

PR #286, opened 2024-10-25
Add traveling-wave cavity model

zihan-zh:
## Description
1. Added a `cavity_type` parameter to the Cheetah Cavity element.
2. Updated the traveling-wave cavity model in the `Cavity._cavity_rmatrix` method to accommodate the new parameter.
3. Modified the Bmad converter function to retrieve the `cavity_type` property.

## Motivation and Context
This change is required to provide users with more versatility when configuring RF cavities in Cheetah.

## Types of changes

## Checklist


jank324:
Just to keep some notes:
 - Removing the vector dimension in `ParticleBeam.from_ocelot` is not the cause of the mismatch between Ocelot and Cheetah in the tests
 - Does the new travelling wave implementation work in a vectorised scenario?

jank324 (approved):

cheetah/converters/ocelot.py
```diff
@@ -116,6 +116,7 @@ def convert_element_to_cheetah(
             voltage=torch.tensor(element.v, dtype=torch.float32) * 1e9,
             frequency=torch.tensor(element.freq, dtype=torch.float32),
             phase=torch.tensor(element.phi, dtype=torch.float32),
+            cavity_type="standing_wave",
             name=element.id,
             device=device,
             dtype=dtype,
```
cr-xu, line 122: I think here we should also include the conversion for `ocelot.TWCacity` as it's implemented now

jank324: @cr-xu like this?

cr-xu: Yes, exactly.

I think the proper test for TWCavity is still missing though. It would be nice to have a benchmark against both OCELOT and Bmad tracking results.

jank324: Yeah. There is a test for the standing wave already that I think can just be parameterised to cover travelling wave as well.

jank324: I just added the test, comparing the results to Ocelot. Weirdly, it seems the `TWCavity` implementation in Ocelot is broken. 🤔

cr-xu: Well... indeed the OCELOT TWCavity seems to be broken, both for the stable `22.12.0` version and the unreleased one in the master branch now `24.12.0`. I already opened an Issue there: https://github.com/ocelot-collab/ocelot/issues/268

Let's see if that can be resolved soon. Otherwise I would suggest to have a static bmad test case and compare the result, so that we can have the PR merged soon-ish.

cr-xu: On a side note, the Ocelot master branch did merge my PR, so now we can also point to its official master instead of my fork for the CI/CD.

jank324: I asked @zihan-zh to provide results from a Bmad tracking, so we can use those for a static comparison.

tests/test_compare_bmad.py
```diff
@@ -155,3 +155,89 @@ def test_different_species_in_different_elements(
...
+    beginning[beta_a] = 5.91253677 ! m a-mode beta function
+    beginning[beta_b] = 5.91253677 ! m b-mode beta function
+    beginning[alpha_a] = 3.55631308 ! a-mode alpha function
+    beginning[alpha_b] = 3.55631308 ! b-mode alpha function
+    beginning[p0c] = 6e6 ! eV    Or can set beginning[p0c]
+
+    parameter[geometry] = open  ! Or closed
+    parameter[particle] = electron  ! Reference particle.
+
+    c: lcavity, rf_frequency = 1.3e9, l = 1.0377, voltage = 0.01815975e9, phi0 = {phase}, cavity_type = {cavity_type}
```
cr-xu, line 178:
```suggestion
    c: lcavity, rf_frequency = 1.3e9, l = 1.0377, voltage = 0.01815975e9, 
        phi0 = {phase/360}, cavity_type = {cavity_type}
```

jank324:
As per #419 and #431, the online Bmad runner was removed again. As a result, we need to do the tests against Bmad with pre-computed saved data again.

jank324:
Only merge this PR after #487 because I made a minor mistake 🙃

jank324:
I've updated the Ocelot tests because `TWCavity is now fully implemented. This seems to mostly work fine, but I'm not sure if the tolerance on `tau` is too high.

tests/test_compare_ocelot.py
```diff
@@ -700,16 +638,25 @@ def test_cavity_non_zero_phase():
...
     # Compare
+    # The 6 particle dimensions are separated to allow for different tolerances
+    assert np.allclose(outgoing_beam.x.cpu().numpy(), outgoing_parray.x(), atol=1e-9)
+    assert np.allclose(outgoing_beam.px.cpu().numpy(), outgoing_parray.px(), atol=1e-4)
+    assert np.allclose(outgoing_beam.y.cpu().numpy(), outgoing_parray.y(), atol=1e-9)
+    assert np.allclose(outgoing_beam.py.cpu().numpy(), outgoing_parray.py(), atol=1e-4)
     assert np.allclose(
-        outgoing_beam.particles[..., :6].cpu().numpy(),
-        outgoing_parray.rparticles.transpose(),
+        outgoing_beam.tau.cpu().numpy(), outgoing_parray.tau(), atol=2e-4
```
jank324, line 654: Is this tolerance too high or is it fine? Currently `2e-4` is more or less the lowest that passes the test.

jank324: @cr-xu?

cr-xu: The 5 and 6-th components are fixed now. There is still some discrepancies in the transverse cooridnates, but I really don't see any difference in the matrix component. As Ocelot now still has "unchecked" comment, we can probably just leave it for future checks.

jank324: Perfect, thank you!

Issue #285, opened 2024-10-25
`plot_reference_particle_traces` crashes for non-leaf tensors


Hespe:
The function `plot_reference_particle_traces` currently crashes if it is called on a `Segment` containing non-leave tensors. The issue is caused by the deepcopy in 

https://github.com/desy-ml/cheetah/blob/01dbf8e26d69039f0c9135c672f88daf649ba962/cheetah/accelerator/segment.py#L438

with error message

> RuntimeError: Only Tensors created explicitly by the user (graph leaves) support the deepcopy protocol at the moment.

Issue #225, opened 2024-07-23
Deepcopy in BPM prevents gradient calculation


Hespe:
In the BPM, there is a call to `deepcopy` before passing on the beam down the line: https://github.com/desy-ml/cheetah/blob/b38a654e0e8971d42d40fad4ffd9f92068140267/cheetah/accelerator/bpm.py#L51

Unfortunetly, this copy currently inhibits taking gradients of the BPM reading. The reduced example

```python
import torch
import cheetah

beam = cheetah.ParameterBeam.from_parameters(mu_x=torch.tensor([0.0], requires_grad=True))

bpm = cheetah.BPM()
bpm.track(beam)
```

fails with

> RuntimeError: Only Tensors created explicitly by the user (graph leaves) support the deepcopy protocol at the moment.  If you were attempting to deepcopy a module, this may be because of a torch.nn.utils.weight_norm usage, see https://github.com/pytorch/pytorch/pull/103001

Hespe:
Maybe the `deepcopy` can simply be replaced by a call to [`torch.clone`](https://pytorch.org/docs/stable/generated/torch.clone.html)?

PR #284, opened 2024-10-22
Improve `Beam` docstrings and type hints

Hespe:
## Description

The outdated docstrings and properties are adjusted to match the new naming scheme.
In addition, `Beam` is converted into an abstract class, increasing the probability to catch such issues early.

(The old `mu_s` property of `Beam` is renamed to `mu_tau`.)

## Motivation and Context

The beam coordinate system has been adjusted recently, leaving some outdated docstrings in the `Beam` class and its subclasses `ParameterBeam` and `ParticleBeam`.

- [x] I have raised an issue to propose this change
  - #238 
  - #279 

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation (update in the documentation)

## Checklist

- [x] I have updated the changelog accordingly (**required**).
- [ ] My change requires a change to the documentation.
- [ ] I have updated the tests accordingly (_required for a bug fix or a new feature_).
- [ ] I have updated the documentation accordingly.
  
  
- [x] I have reformatted the code and checked that formatting passes (**required**).
- [x] I have have fixed all issues found by `flake8` (**required**).
- [x] I have ensured that all `pytest` tests pass (**required**).
- [ ] I have run `pytest` on a machine with a CUDA GPU and made sure all tests pass (**required**).
- [x] I have checked that the documentation builds (**required**).

Note: We are using a maximum length of 88 characters per line.

PR #289, opened 2024-10-28
Add `clone` method for beams and elements

Hespe:
## Description

There is currently no reliable way to obtain copies of `Beam` or `Element`. This has lead to `copy.deepcopy` being used in some places which is troublesome in some instances (see the issues linked below).

## Motivation and Context

- [x] I have raised an issue to propose this changs
  - #225
  - #285

## Types of changes

## Checklist


Hespe:
Would be wise to merge this PR after #284 such that `Beam.clone()` can be marked abstract.

Hespe:
The second test failure should be fixed once #288 is merged.

Hespe:
The second test failure should be fixed once #288 is merged.

Hespe:
@jank324 There are some lines in the plotting tests that change context beyond their own scope:

https://github.com/desy-ml/cheetah/blob/01dbf8e26d69039f0c9135c672f88daf649ba962/tests/test_plotting.py#L16-L20

Not sure if we should adjust that somehow or if its fine in this case.

Hespe:
> Looks like the `TransverseDeflectingCavity` is broken if its used within a `Segment` and not on its own. The problem seems to be that `Segment.track()` calls `super.track()`, that is `Element.track()`, if the `Segment` is skippable. However, `Element.track()` falls back to the transfer map, which is not implemented for the cavity.

Raised this issue in #290

jank324:
#288 is merged now.

Hespe:
The plotting related failure is now fixed. This leaves only #290 as a blocker for this PR.

cheetah/accelerator/rbend.py
```diff
@@ -51,6 +55,7 @@ def __init__(
             fringe_integral=fringe_integral,
             fringe_integral_exit=fringe_integral_exit,
             gap=gap,
+            gap_exit=gap_exit,
             name=name,
             device=device,
             dtype=dtype,
```
jank324, line 61: @cr-xu we are struggling with this PR a bit and I think you might have useful input.

In `RBend` there might be an issue with `e1` and `e2`. If I instantiate an `RBend` and then access `my_rbend.e1`, the `e1` I get is different from the one I set in the instantiation, by being in `Dipole` space rather than `RBend` space. That's an issue, right?

@Hespe and I were brainstorming solutions and one idea was to define two different `e1` and `e2` ... something like `dipole_e1` and `rbend_e1`, where `rbend_e1` would just be a computed property of the `RBend` class that computes based on `dipole_e1`.

What are your thoughts on this? Are you aware of any conventions already present in other codes or in general notation?

Hespe:
I think the generic `clone()` implementation will not work for the `CustomTransferMap`. The problem is that the important property is `_transfer_map` but the constructor parameter does not have the underscore in front.

jank324:
> I think the generic `clone()` implementation will not work for the `CustomTransferMap`. The problem is that the important property is `_transfer_map` but the constructor parameter does not have the underscore in front.

Ugh ... yeah. I guess we either add a custom implementation for this special case, or rename the value of the transfer map to distinguish it from the function that every element has.

Hespe:
One more thing I've noticed: `name` does not seem to be part of `defining_features` for any of the elements. Cloning would therefore remove all name information at the moment.

cheetah/accelerator/bpm.py
```diff
@@ -62,7 +61,7 @@ def plot(self, ax: plt.Axes, s: float, vector_idx: Optional[tuple] = None) -> No
 
     @property
     def defining_features(self) -> list[str]:
-        return super().defining_features
+        return super().defining_features + ["is_active"]
```
jank324, line 64: For elements where `is_active` can be set manually, shouldn't it be part of `defining_properties`, i.e. wouldn't we want it to also be included in when lattices are saved as well?

jank324:
> One more thing I've noticed: `name` does not seem to be part of `defining_features` for any of the elements. Cloning would therefore remove all name information at the moment.

It probably makes sense for that to be a defining feature of the `Element` base class already, right?

cheetah/accelerator/bpm.py
Hespe: I had thought that I applied that change a few minutes ago. Maybe something went wrong with that

Hespe: Ah, I misunderstood. Is your suggestion to apply the same change to other elements with a manually set `is_active` property?

jank324:
@cr-xu Does this PR now also fix #298?

jank324:
@Hespe I can't officially request a review, so here is an unofficial request.

cr-xu:
> @cr-xu Does this PR now also fix #298?

Feature / method-wise yes. 
I think we should make it clear in the documentation how these values are handled, but this can be done afterwards.

jank324 (approved):

cheetah/particles/particle_beam.py
```diff
@@ -1074,6 +1074,13 @@ def momenta(self) -> torch.Tensor:
         """Momenta of the individual particles."""
         return torch.sqrt(self.energies**2 - electron_mass_eV**2)
 
+    def clone(self) -> "ParticleBeam":
+        return ParticleBeam(
+            particles=self.particles.clone(),
+            energy=self.energy.clone(),
+            particle_charges=self.particle_charges.clone(),
+        )
```
Hespe, line 1082: I suppose the survival probability is missing from this list? We merged the other PR here after I had written the `clone` method.

tests/test_clone.py
```diff
@@ -0,0 +1,40 @@
...
+
+    for buffer, buffer_clone in zip(element.buffers(), clone.buffers()):
+        assert torch.allclose(buffer, buffer_clone)
+        assert not buffer.data_ptr() == buffer_clone.data_ptr()
+
+
+@pytest.mark.parametrize("BeamClass", [cheetah.ParameterBeam, cheetah.ParticleBeam])
+def test_beam_buffer_location(BeamClass):
+    """Test that the buffers of clones beams do not share memory."""
+    beam = BeamClass.from_parameters()
```
Hespe, line 35: Initializing the beam with the standard arguments might be a bad idea now that I think about it. If a property is absent from `clone`, it will be initialized by its default and we will not catch that problem here.

Hespe:
Looks good to me except for these two comments about potentially missing features in the `clone` implementation.

tests/test_clone.py
jank324: Hmm ... maybe the test in general is a bit inconsistent. The name and docstring suggest that it only tests buffer location (where the defaults wouldn't be an issue I think), but it also checks the values

cheetah/accelerator/dipole.py
cr-xu, line 1: The docstring for dipole also needs to be changed e1, e2 -> dipole_e1, dipole_e2

cheetah/accelerator/rbend.py
```diff
@@ -17,11 +17,21 @@ class RBend(Dipole):
     :param k1: Focussing strength in 1/m^-2.
     :param e1: The angle of inclination of the entrance face [rad].
     :param e2: The angle of inclination of the exit face [rad].
```
cr-xu, line 19:
```suggestion
    :param rbend_e1: The angle of inclination of the entrance face in rad.
    :param rbend_e2: The angle of inclination of the exit face in rad.
```

cr-xu:
Apart from that, it's good to go from my side.

PR #292, opened 2024-11-06
Distribution plotting

roussel-ryan:
## Description

Adds plotting functionality to the `ParticleBeam` class allowing for 2D projection visualization of up to 6D phase spaces.

Implements the method `plot_distribution` in the `ParticleBeam` class to generate `matplotlib` plots of the 6D phase space distributions. Optional arguments can customize which phase spaces to plot, specify custom bin limits, number of bins per axis, etc. 

## Motivation and Context

This method allows for easy visualization of the phase space distribution.

## Types of changes

## Checklist


jank324:
For reference, this PR now adds convenience functions for four plots of `ParticleBeam`.

Assuming
```python
beam = cheetah.ParticleBeam.from_astra("../tests/resources/ACHIP_EA1_2021.1351.001")
```

You can create a 1D distribution plot
```python
beam.plot_1d_distribution("p", bins=300)
```
Screenshot 2024-12-09 at 17 48 05

a 2D distribution plot
```python
beam.plot_2d_distribution("x", "px", bins=100)
```
Screenshot 2024-12-09 at 17 48 45

... even with contour lines if desired
```python
beam.plot_2d_distribution("x", "px", contour=True)
```
Screenshot 2024-12-09 at 17 50 09

A combined distribution plot of the entire beam (following the one by @roussel-ryan)
```python
beam.plot_distribution(bin_ranges="unit_same")
```
Screenshot 2024-12-09 at 17 50 35

as well as a point cloud plot
```python
beam.plot_point_cloud(scatter_kws={"s": 1, "alpha": 0.05})
```
Screenshot 2024-12-09 at 17 52 05

Everything follows the conventions of `seaborn` for providing pre-made `matplotlib` plots, meaning it is endlessly modifiable through additional keywoard argument dictionaries and should feel familiar to anyone who already uses the ecosystem around `matplotlib`.

jank324:
Maybe a small note here: Right now the histograms all assume particle counts. It is imaginable that someone might want charge or mass as a weight for each particle. Right now all particles have the same charge or mass, so I wouldn't solve this problem now, but we might add this as an option at some point.

cheetah/particles/particle_beam.py
```diff
@@ -883,6 +898,351 @@ def to_xyz_pxpypz(self) -> torch.Tensor:
...
+        )
+        ax.set_xlabel(f"{self.PRETTY_DIMENSION_LABELS[dimension]}")
+
+        # Handle units
+        if dimension in ("x", "y", "tau"):
+            base_unit = "m"
+        elif dimension in ("px", "py"):
+            base_unit = "rad"
+        elif dimension == "p":
+            base_unit = "%"
```
cr-xu, line 947: I think the proper way to say is $px, py, p$ are all dimensionless. 

- px,py are not exactly the geometrical slope, so using $rad$ might be misleading
- p is just a ratio, if expressed in % it still needs to be multiplied by 100

jank324: Ah ... I didn't realise that about $px$ and $py$. With $p$, the tick formatter does the multiplication by 100 ... I'm not sure if the variable name `base_unit` is a bit misleading here in that case (it was just `unit` before).

So you would propose to do all three in % then?

cr-xu: I would suggest to put them as dimensionless (no unit), if that's possible.

jank324: So kind of how I changed it now?

PR #301, opened 2024-12-06
Add test for using `nn.Parameter` as element length

### LINKED ISSUE #164: length cannot be a NN parameter
I have been unable to let length be an optimizable Pytorch.nn.Parameter without corrupting the transfer map.  I think it is because every element inherits a length from the Element class. 
https://github.com/desy-ml/cheetah/blob/3fe16a797ccb141abd2c13be4342a693fe0b2585/cheetah/accelerator/element.py#L32

 If I comment this out, then I am able to assign length as a Parameter, but this then causes some tests to fail (I believe it is every CI test using transfer maps).  

I haven't found a good fix yet but it would be nice to allow length to be a parameter

Hespe:
## Description

It was previously observed that `Element.length` cannot be set to `torch.nn.Parameter` instead of `torch.Tensor`. However, up-to-date version of Cheetah do already support such an assignment.

It is likely that the required change was introduced in 383ff83. To prevent the issue from reappearing, we implement a simple test case.

## Motivation and Context

  - Closes #164

## Types of changes

## Checklist

### DIFF
--- a/tests/test_drift.py
+++ b/tests/test_drift.py
@@ -91,10 +91,8 @@ def test_drift_bmadx_tracking(dtype):
 
 
 def test_length_as_parameter():
-    """
-    Test that the drift length can be set as a torch.nn.Parameter.
-    """
-    length = torch.Tensor([1.0])
+    """Test that the drift length can be set as a `torch.nn.Parameter`."""
+    length = torch.Tensor(1.0)
     parameter = torch.nn.Parameter(length)
 
     # Create to equal drifts, one with Tensor, one with Parameter
--- a/tests/test_drift.py
+++ b/tests/test_drift.py
@@ -92,7 +92,7 @@ def test_drift_bmadx_tracking(dtype):
 
 def test_length_as_parameter():
     """Test that the drift length can be set as a `torch.nn.Parameter`."""
-    length = torch.Tensor(1.0)
+    length = torch.tensor(1.0)
     parameter = torch.nn.Parameter(length)
 
     # Create to equal drifts, one with Tensor, one with Parameter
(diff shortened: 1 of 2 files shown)


tests/test_drift.py
```diff
@@ -88,3 +88,25 @@ def test_drift_bmadx_tracking(dtype):
         atol=1e-14 if dtype == torch.float64 else 0.00001,
         rtol=1e-14 if dtype == torch.float64 else 1e-6,
     )
+
+
+def test_length_as_parameter():
+    """
+    Test that the drift length can be set as a torch.nn.Parameter.
+    """
```
jank324, line 96:
```suggestion
    """Test that the drift length can be set as a `torch.nn.Parameter`."""
```

tests/test_drift.py
```diff
@@ -88,3 +88,25 @@ def test_drift_bmadx_tracking(dtype):
...
+    length = torch.Tensor([1.0])
```
jank324, line 97:
```suggestion
    length = torch.Tensor(1.0)
```

jank324 (approved):

tests/test_drift.py
```diff
@@ -88,3 +88,23 @@ def test_drift_bmadx_tracking(dtype):
         atol=1e-14 if dtype == torch.float64 else 0.00001,
         rtol=1e-14 if dtype == torch.float64 else 1e-6,
     )
+
+
+def test_length_as_parameter():
+    """Test that the drift length can be set as a `torch.nn.Parameter`."""
+    length = torch.Tensor(1.0)
```
jank324, line 95:
```suggestion
    length = torch.tensor(1.0)
```

jank324:
The content of this PR is probably fine, but there is a test dependency that has disappeared for a version of Ocelot. This probably just has to be linked back to the original Ocelot repository and everything will be fine again.

Hespe:
Seems to have been a spurios issue on Github's side. Re-running the tests was enough for it not to appear again.

jank324:
> Seems to have been a spurios issue on Github's side. Re-running the tests was enough for it not to appear again.

Odd ...

PR #303, opened 2024-12-09
Fix `nn.Parameter` assignment at initialization

### LINKED ISSUE #302: `nn.Parameter` provided at initialization does not appear in `Element.parameters`
`torch.nn.Parameters` that are provided to the constructor of Cheetah elements are not properly included in the list of parameters provided by `Element.parameters()`. If instead the parameter is set manually after creating the element, the parameter is properly registered.

Hespe:
## Description

Currently, a `torch.nn.Parameter` provided at initialization to Cheetah elements does not appear in the list of parameters returned by `.parameters()`, in contrast to a parameter assigned after initialization. This PR makes the behaviour consistent.

## Motivation and Context

  - Closes #302 

## Types of changes

## Checklist

### DIFF
--- a/cheetah/particles/parameter_beam.py
+++ b/cheetah/particles/parameter_beam.py
@@ -35,17 +35,16 @@ class ParameterBeam(Beam):
         factory_kwargs = {"device": device, "dtype": dtype}
         super().__init__()
 
-        self.register_buffer("_mu", torch.as_tensor(mu, **factory_kwargs))
-        self.register_buffer("_cov", torch.as_tensor(cov, **factory_kwargs))
-        total_charge = (
-            total_charge
-            if total_charge is not None
-            else torch.tensor([0.0], **factory_kwargs)
-        )
-        self.register_buffer(
-            "total_charge", torch.as_tensor(total_charge, **factory_kwargs)
-        )
-        self.register_buffer("energy", torch.as_tensor(energy, **factory_kwargs))
+        self.register_buffer("_mu", None)
+        self.register_buffer("_cov", None)
+        self.register_buffer("energy", None)
+        self.register_buffer("total_charge", torch.tensor(0.0, **factory_kwargs))
+
+        self._mu = torch.as_tensor(mu, **factory_kwargs)
+        self._cov = torch.as_tensor(cov, **factory_kwargs)
+        self.energy = torch.as_tensor(energy, **factory_kwargs)
+        if total_charge is not None:
+            self.total_charge = torch.as_tensor(total_charge, **factory_kwargs)
 
     @classmethod
     def from_parameters(
--- a/cheetah/particles/particle_beam.py
+++ b/cheetah/particles/particle_beam.py
@@ -54,24 +54,21 @@ class ParticleBeam(Beam):
             particles.shape[-2] > 0 and particles.shape[-1] == 7
         ), "Particle vectors must be 7-dimensional."
 
-        self.register_buffer("particles", particles.to(**factory_kwargs))
+        self.register_buffer("particles", None)
+        self.register_buffer("energy", None)
         self.register_buffer(
-            "particle_charges",
-            (
-                particle_charges.to(**factory_kwargs)
-                if particle_charges is not None
-                else torch.zeros(particles.shape[-2], **factory_kwargs)
-            ),
+            "particle_charges", torch.zeros(particles.shape[-2], **factory_kwargs)
         )
-        self.register_buffer("energy", energy.to(**factory_kwargs))
         self.register_buffer(
-            "survival_probabilities",
-            (
-                survival_probabilities.to(**factory_kwargs)
-                if survival_probabilities is not None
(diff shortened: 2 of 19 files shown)


cheetah/particles/particle_beam.py
```diff
@@ -54,25 +54,22 @@ def __init__(
...
+            "survival_probabilities", torch.ones(particles.shape[-2], **factory_kwargs)
         )
 
+        self.particles = particles.to(**factory_kwargs)
+        self.energy = energy.to(**factory_kwargs)
+        if particle_charges is not None:
+            self.particle_charges = particle_charges.to(**factory_kwargs)
+        if survival_probabilities is not None:
+            self.survival_probabilities = survival_probabilities.to(**factory_kwargs)
+
```
jank324, line 72: Same here (?)

cheetah/particles/parameter_beam.py
```diff
@@ -35,17 +35,16 @@ def __init__(
...
+        self.register_buffer("_mu", None)
+        self.register_buffer("_cov", None)
+        self.register_buffer("energy", None)
+        self.register_buffer("total_charge", torch.tensor(0.0, **factory_kwargs))
+
+        self._mu = torch.as_tensor(mu, **factory_kwargs)
+        self._cov = torch.as_tensor(cov, **factory_kwargs)
+        self.energy = torch.as_tensor(energy, **factory_kwargs)
+        if total_charge is not None:
+            self.total_charge = torch.as_tensor(total_charge, **factory_kwargs)
```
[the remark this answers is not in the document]
Hespe: When you register a buffer and pass a `torch.nn.Parameter` for initialization, the parameter is not properly registered and does not appear in `Module.parameters()` but `Module.buffers()`. Therefore I switched to a late initialization to make `torch.nn.Module` handle the parameter registration properly.
Passing `None` is necessary since the second argument is mandatory and does not have a default.

An alternative would be to check whether the passed arguments are `torch.Tensor` or `torch.nn.Paramter` and then call either `register_buffer()` or `register_parameter` appropriately.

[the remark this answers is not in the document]
jank324: Ah, I think I got confused ... so
```python
self.register_buffer("_mu", torch.as_tensor(mu, **factory_kwargs))
self.register_buffer("_cov", torch.as_tensor(cov, **factory_kwargs))
self.register_buffer("energy", torch.as_tensor(energy, **factory_kwargs))
self.register_buffer("total_charge", torch.as_tensor(total_charge, **factory_kwargs))
```
would basically produce the original bug, right?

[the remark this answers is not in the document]
Hespe: Exactly. But I think we can actually do without manual registration for the arguments without default.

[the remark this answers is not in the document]
Hespe: Ok, no. We do need to manually register the buffers. Only `nn.Parameter` is registered automatically.

Hespe:
Hmm, I don't get it. Did we just get super unlucky and the test failed twice in a row based on a bad particle distribution? This part of the code should not be affected at all by the last commit.

jank324:
> Hmm, I don't get it. Did we just get super unlucky and the test failed twice in a row based on a bad particle distribution? This part of the code should not be affected at all by the last commit.

That appears to be the case.

tests/test_dipole.py
```diff
@@ -155,3 +155,17 @@ def test_dipole_bmadx_tracking(dtype):
...
+    Test that buffers are properly registered in the dipole element.
+    """
+    length = torch.tensor(0.5)
+    gap_exit = torch.tensor(0.1)
+    dipole = Dipole(length=length, gap_exit=gap_exit)
+
+    # length and gap_exit are tested because they behave slightly different than other
+    # properties. length is inherited and gap_exit is derived from gap by default.
+    assert length in dipole.buffers()
+    assert gap_exit in dipole.buffers()
```
jank324, line 171: I'm a little confused by this test. It only tests the "weird" buffers? Then this should be reflected in the docstring and name (I guess by saying in the docstring what it says in the comment).

Hespe: In principle I would have liked to test all buffers, but as we said last week there does not seem to be a good way to achieve this automatically. Therefore I selected the two arguments that seemed most likely to cause trouble to me. We could of course extend this to cover more buffers of `Dipole`.

jank324: @Hespe, what do you think about the `test_buffer_registration` test now? I just added all buffers. I think because the test is purely for `Dipole` it's fair to handcode what we would expect.

Hespe: Seems reasonable. But its something to keep in mind should we need to add more arguments/buffers later.

jank324: Okay, I've added a check for the number of buffers. If we add stuff, there is a good chance this will change and alert us to update the test.

jank324 (approved):

PR #314, opened 2024-12-20
Allow passing `float` for scalar values

### LINKED ISSUE #313: Allow passing Python `float` instead of `torch.Tensor` for scalar arguments
Many methods currently assumed that a `torch.Tensor` is passed for their arguments even for scalar values. This makes calling them more verbose than necessary. Instead, it would be nice if `float` could be passed instead.

Hespe:
## Motivation and Context

  - Closes #313 

## Types of changes

## Checklist

### DIFF
--- a/cheetah/particles/parameter_beam.py
+++ b/cheetah/particles/parameter_beam.py
@@ -92,36 +92,78 @@ class ParameterBeam(Beam):
         factory_kwargs = {"device": device, "dtype": dtype}
 
         # Set default values without function call in function signature
-        mu_x = mu_x if mu_x is not None else torch.tensor(0.0, **factory_kwargs)
-        mu_px = mu_px if mu_px is not None else torch.tensor(0.0, **factory_kwargs)
-        mu_y = mu_y if mu_y is not None else torch.tensor(0.0, **factory_kwargs)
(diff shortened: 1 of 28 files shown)


jank324:
Yeah, I'm not sure that we will merge this PR though. I also like the idea of only having to pass a float, but there are some arguments against doing this:
 - For consistency, if you can do
```python
quad = cheetah.Quadrupole(k1=4.2)
```
you should also be able to do
```python
quad.k1 = 4.2
```
This would lead to a lot of boiler plate code inside of Cheetah that wouldn't be nice and could also slow it down a little.
 - Doing this would also result in an incosistency because
```python
my_k1 = 4.2
quad.k1 = my_k1
quad_k1 == my_k1   # Would be False even though you would expect True
```
 - Overall, this feels a little like it would go against the Zen of Python _"Explicit is better than implicit"_

That said, we don't endorse it, but in most cases I believe you can just pass floats to inits in the current version of Cheetah :thinking:

Hespe:
Just to point that out, passing floats is currently prevented by `verify_device_and_dtype`. So an option might actually be to reduce this PR to adjusting `verify_device_and_dtype` for `float` inputs.

PR #340, opened 2025-02-11
Automatically detect default `dtype` for lattice and beam importers

### LINKED ISSUE #333: Adapt automatic type scheme for importers
When creating Cheetah beams and elements, the dtype is automatically determined in correct torch-style. This is currently not done in beam and lattice importers ... but it should be.

Hespe:
## Motivation and Context

  - Closes #333 

## Types of changes

## Checklist

### DIFF
--- a/cheetah/converters/astra.py
+++ b/cheetah/converters/astra.py
@@ -1,5 +1,3 @@
-from typing import Tuple
-
 import numpy as np
 from scipy.constants import physical_constants
 
@@ -7,7 +5,7 @@ from scipy.constants import physical_constants
 electron_mass_eV = physical_constants["electron mass energy equivalent in MeV"][0] * 1e6
 
 
-def from_astrabeam(path: str) -> Tuple[np.ndarray, float, np.ndarray]:
+def from_astrabeam(path: str) -> tuple[np.ndarray, float, np.ndarray]:
     """
     Read from a ASTRA beam distribution, and prepare for conversion to a Cheetah
     ParticleBeam or ParameterBeam.
--- a/cheetah/converters/astra.py
+++ b/cheetah/converters/astra.py
@@ -39,7 +39,7 @@ def from_astrabeam(path: str) -> tuple[np.ndarray, float, np.ndarray]:
     energy = gamref * electron_mass_eV
 
     n_particles = xp.shape[0]
-    particles = np.zeros((n_particles, 6), dtype=P0.dtype)
+    particles = np.zeros((n_particles, 6))
 
     u = np.c_[xp[:, 3], xp[:, 4], xp[:, 5] + Pref]
     gamma = np.sqrt(1 + np.sum(u * u, 1) / electron_mass_eV**2)
--- a/cheetah/converters/astra.py
+++ b/cheetah/converters/astra.py
@@ -1,3 +1,5 @@
+from typing import Tuple
+
 import numpy as np
 from scipy.constants import physical_constants
 
@@ -5,7 +7,7 @@ from scipy.constants import physical_constants
 electron_mass_eV = physical_constants["electron mass energy equivalent in MeV"][0] * 1e6
 
 
-def from_astrabeam(path: str) -> tuple[np.ndarray, float, np.ndarray]:
+def from_astrabeam(path: str) -> Tuple[np.ndarray, float, np.ndarray]:
     """
     Read from a ASTRA beam distribution, and prepare for conversion to a Cheetah
     ParticleBeam or ParameterBeam.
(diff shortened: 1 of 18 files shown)


Hespe:
Before finishing this PR, we should discuss the default behaviour of the beam importers that import numpy data. Right now there are actually in an inconsistent state:

- When `from_ocelot` is used with default `dtype=None`, it uses the `dtype` of the numpy array passed to it. In my opinion this is the correct behaviour and it mimickes the `dtype` one obtains from `torch.tensor` when passing a numpy array. In fact, the behaviour is implemented using `torch.as_tensor` to determine the appropriate `dtype`.
- `from_astra` on the other hand uses a default of `dtype=torch.float32`. This does not follow the automatic `dtype` selection we are implementing in this PR and thus definitely has to change. I see two possibilities:
  1. Simply change to `dtype=None`, making its behaviour identical to `from_ocelot`. This would be my preferred choice but it leads to a number of test failures currently. The problem is that the tests implicitly assume a beam of type `torch.float32` and the solution would be to explicitly set `dtype=torch.float32` in the tests while loading the beam.
  2. Change to `dtype=None`, but query `torch.get_default_dtype` in that case to determine the `dtype`. This makes the API mostly consistent with the current behaviour since `torch.get_default_dtype()` defaults to `torch.float32` but deviates from how `torch.tensor` determines its `dtype` from numpy arrays.

In case we go for the second option, `from_ocelot` should be updated in the same way.

cr-xu:
Hi, I just noticed that the lattice loaded with `Segment.from_lattice_json()` also doesn't have an option to select dtype, which is annoying because it now defaults to float64.

cr-xu:
On a side note, I discovered that now the entire `Segment` can be reliably casted to float or double simply by calling
`segment.to(torch.float32)`
Did we know that already? @Hespe @jank324  I vaguely remembered that we were discussing this at some point but it was not possible. Probably it's worth to document it somewhere.

Hespe:
You should even be ablo to simply call `segment.double()` or `segment.single()` to convert between `float32` and `float64`, same for instances of `Beam`. However, you are right that this should be properly documented somewhere.

Hespe:
> Hi, I just noticed that the lattice loaded with `Segment.from_lattice_json()` also doesn't have an option to select dtype, which is annoying because it now defaults to float64.

I've just added a `device` and `dtype` argument to `Segment.from_lattice_json()`. Furthermore, it should now always default to `torch.float32` unless you use `torch.set_default_dtype`. But to be fair, I'm not even sure in which cases it would previously default to`torch.float64`.

cr-xu:
> > Hi, I just noticed that the lattice loaded with `Segment.from_lattice_json()` also doesn't have an option to select dtype, which is annoying because it now defaults to float64.
> 
> I've just added a `device` and `dtype` argument to `Segment.from_lattice_json()`. Furthermore, it should now always default to `torch.float32` unless you use `torch.set_default_dtype`. But to be fair, I'm not even sure in which cases it would previously default to`torch.float64`.

Thanks! I was testing that on released v0.7.0 version, not this PR branch.

Hespe:
I noticed a small inconsistency in the `dtype` of cheetah elements while writing this PR. All elements inherit a `length` property from `Element`. But for elements that are not supposed to have a length, e.g., `Marker`, there is no `length` argument to their constructors and also no `dtype` and `device` arguments such that their `length` always has `dtype` `torch.float32`. Importing or constructing lattices of type `torch.float64` therefore leads to a `Segment` with mixed `dtype`.

cheetah/converters/astra.py
```diff
@@ -1,11 +1,13 @@
...
+
 import numpy as np
 from scipy.constants import physical_constants
 
 # Electron mass in eV
 electron_mass_eV = physical_constants["electron mass energy equivalent in MeV"][0] * 1e6
 
 
-def from_astrabeam(path: str) -> tuple[np.ndarray, float, np.ndarray]:
+def from_astrabeam(path: str) -> Tuple[np.ndarray, float, np.ndarray]:
```
jank324, line 10: I'm pretty sure that non-capital `tuple` is the more modern type annotation because it doesn't require the import.

cheetah/latticejson.py
```diff
@@ -131,64 +131,103 @@ def encode(self, obj, level=0):
...
 
     :param value: Value to convert to a `torch.Tensor` if necessary.
+    :param device: Device to place the lattice elements on.
+    :param dtype: Data type to use for the lattice elements.
     :return: Value converted to a `torch.Tensor` if necessary.
     """
-    return value if isinstance(value, (str, bool)) else torch.tensor(value)
+    return (
+        value
+        if isinstance(value, (str, bool))
```
jank324, line 151:
```suggestion
        if isinstance(value, (str, bool, int))
```
@Hespe ... this feels like the change that should fix the resolution issue, because no tensor value in Cheetah should ever be an int ... but I tested it and it doesn't fix the issue.

This raises the question: Does JSON distinguish between int and float?

**NOTE:** If we make this change, the docstring needs to be changed accordingly.

jank324 (approved):

jank324 (changes requested):
Changelog has not bred added yet

Hespe:
> @Hespe ... I believe I fixed the LatticeJSON screen resolution issue no in a way that actually even matches the mechanisms we already had in place. Can you check the solution though?
> 
> Otherwise, I think this PR is ready.

Looks good to me. The only thing that comes to my mind would be if some property has value e.g. `torch.tensor([1.0, 2.0])` if that would be reimported as `list[int]` instead of `list[float]` and thus not cast to a tensor.

jank324 (approved):

PR #352, opened 2025-02-24
 Integration of Lattice Segment Visualization Tool

chrisjcc:
## Description

This update integrates a 3D visualization feature for lattice segments, enabling users to generate and explore the geometric configurations of lattice structures defined in a JSON configuration file. By executing the command view-lattice --config config.json, the tool processes the lattice segment data from the provided configuration file and produces a corresponding 3D model. The resulting visualization can be interactively explored, providing an intuitive representation of the segment’s geometry.

## Motivation and Context

The ability to visualize lattice segments in 3D is crucial for better understanding their spatial configuration and the relationships between different segments. This feature facilitates the inspection and validation of lattice geometries defined in simulation settings or experimental configurations. The integration of 3D visualization directly addresses the need for a more interactive and intuitive interface for users working with complex lattice models.

## Types of changes

- [ x] New feature (non-breaking change which adds functionality)

## Checklist

- [x ] My change requires a change to the documentation.
- [x ] I have updated the tests accordingly (_required for a bug fix or a new feature_).
- [x ] I have updated the documentation accordingly.
  
  
- [x ] I have reformatted the code and checked that formatting passes (**required**).
- [x ] I have fixed all issues found by `flake8` (**required**).


jank324:
Seems to scale to the full ARES lattice too. But the `trimesh` viewer becomes unusably slow. The external viewer runs perfectly. The full lattice highlights the importance of scaling the lengths of the elements according to their `length` property.

Screenshot 2025-07-02 at 16 37 31

chrisjcc:
@jank324, I don't recall the rendering being particularly slow. Off the top of my head, I can't say what might be causing it—it's been quite a while since I last ran the code. I hadn't noticed these test formats before, but I thought I could help troubleshoot in case it was related to import issues. Cool to see the scale of the full ARES lattice!

chrisjcc:
@jank324, I reinstalled the code locally to refresh my memory on how the trimesh rendering behaved. I now recall that the default camera angle differed between Blender, the trimesh render, and the GLB viewer. This made it tricky at first to figure out each one's default axis orientation, what’s considered “up,” “forward,” etc., and to decide which one felt most intuitive or correct. I ended up using Blender’s axis convention as the canonical reference (mention for historical context). I also remember that the default orientation in trimesh wasn’t particularly ideal. That said, I didn’t experience any slowdown when zooming or moving the camera around on my end (perhaps machine-dependent, not sure), tested based on the small example code you provided above.

cheetah/utils/warnings.py
```diff
@@ -53,3 +53,12 @@ class DirtyNameWarning(Warning):
...
     ...
+
+
+class NoVisualizationWarning(Warning):
+    """
+    Warning raised when an element does not have a visualization method, i.e. it cannot
+    be converted to a 3D mesh.
+    """
+
+    ...
```
jank324, line 64: @Hespe, this might be something we could reuse for `Element.plot`?

jank324:
I simplified the way this feature is used now, and added a feature that scales the 3D model for each element according to its length. The new way to use the 3D visualisation now is the following

```python
import cheetah
import torch
import trimesh


segment = cheetah.Segment(
    elements=[
        cheetah.Cavity(length=torch.tensor(1.141)),
        cheetah.Drift(length=torch.tensor(0.175)),
        cheetah.Quadrupole(length=torch.tensor(0.122), name="AREAMQZM1"),
        cheetah.Drift(length=torch.tensor(0.428)),
        cheetah.Quadrupole(length=torch.tensor(0.122), name="AREAMQZM2"),
        cheetah.Drift(length=torch.tensor(0.204)),
        cheetah.VerticalCorrector(length=torch.tensor(0.02), name="AREAMCVM1"),
        cheetah.Drift(length=torch.tensor(0.204)),
        cheetah.Quadrupole(length=torch.tensor(0.122), name="AREAMQZM3"),
        cheetah.Drift(length=torch.tensor(0.179)),
        cheetah.HorizontalCorrector(length=torch.tensor(0.02), name="AREAMCHM1"),
        cheetah.Drift(length=torch.tensor(0.45)),
        cheetah.Screen(name="AREABSCR1"),
    ]
)

mesh = segment.to_mesh(s=0.0)

mesh.show()
```

Screenshot 2025-07-02 at 18 22 43

And you can save it to view it elsewhere.

```python
mesh.export(file_obj="mesh.glb", file_type="glb")
```

Screenshot 2025-07-02 at 18 23 46

Of course the camera is way too large. I think the model in the file should just be scaled down rather than doing this in code. The reason it's so large is because it has 0.0 length and those models are not scaled but added as they are in the file.

@chrisjcc considering that you say the default texture is glossy, I think it throws out the textures somewhere. I don't understand where, because the thing the code did between it working and to working was technically identical.

I would like to preserve the original textures, because GLB seems to match them nicely to what I originally specced in Blender. A fix should therefore go in that direction.

jank324:
Full ARES looks better now as well ... if we ignore those huge cameras 🙈

Screenshot 2025-07-02 at 18 33 13

chrisjcc:
> The blue tint of the camera is also another remaining issue.
I believe this issue needs to be addressed within Blender, specifically related to the camera artifact. For some reason, the camera screen was rendering black instead of the intended greyish color. Since the background was also black, the screen ended up blending into it. It looked correct in the Blender editor but not in the final compiled asset file, and I couldn’t figure out why. As a quick workaround, I changed the screen color to a bluish tone. However, since the camera asset was linked to the diagnostic screen, it also inherited the new color. Unfortunately, I ran out of time to find a better solution.

cheetah/utils/warnings.py
Hespe: The way to go would be to add a default implementation of `plot` in `Element` which does not plot but throws a warning? Not sure, in principle `@abstractmethod` should stop you from even instanciating a subclass which does not override `plot`.

jank324:
Btw ... I ran the steps that the code does to build the model manually on a single element, and didn't get the glossiness. Something is fishy here. 🐟

jank324:
To be discussed, but I think now that the 3D models should be a separate repository that can also be separately citable.

jank324:
The history of this too screwed up, and trying to fix it, I made it worse. 

I will replay the changes on #502, which will move forward towards the merge. **This PR is therefore closed as abandoned.**

PR #360, opened 2025-02-27
Update to new style type annotations (post 3.9)

jank324:
## Description

Changes type annotations from old to new style, which was introduced in Python 3.10.

Example 1: `Optional[float]` becomes `float | None`
Example 2 `Union[float, str]` becomes `float | str`
Example 3: `Tuple` becomes `tuple`

It is worth noting that while this is nicer (and probably saves a little bit of import time) this PR will definitely break Python 3.9 compatibility. Technically this is fine though because we no longer support 3.9.

## Motivation and Context

Gently recommended by Python core developers. Also usually more readable.

## Types of changes

## Checklist


jank324:
I noticed while doing this that some beam properties could be `None` according to the type annotations, but that doesn't make sense. Probably something for #359.

cr-xu:
Also the newly merged `species.py` needs to be reformatted. Other than that it looks good to me.

cheetah/particles/particle_beam.py
```diff
@@ -980,10 +985,10 @@ def from_xyz_pxpypz(
...
         energy: torch.Tensor,
-        particle_charges: Optional[torch.Tensor] = None,
-        survival_probabilities: Optional[torch.Tensor] = None,
-        species: Optional[Species] = None,
-        device=None,
+        particle_charges: torch.Tensor | None = None,
+        survival_probabilities: torch.Tensor | None = None,
+        species: Species | None = None,
+        device: torch.device | None = None,
         dtype=torch.float32,
```
Hespe, line 992: We seems to have missed this converter function in the recent PR.

cheetah/particles/particle_beam.py
```diff
@@ -1404,15 +1407,15 @@ def num_particles_survived(self) -> torch.Tensor:
         return self.survival_probabilities.sum(dim=-1)
 
     @property
-    def x(self) -> Optional[torch.Tensor]:
+    def x(self) -> torch.Tensor | None:
```
Hespe, line 1410: Is there a reason for having all those properties as Optional?

jank324: No ... that's what I was talking about one Friday. I wasn't sure weather to already fix this as part of this PR ... but now I'm thinking yes.

I don't see how that (and the other similar) computations would result in `None`, so I guess these should be corrected.

cheetah/particles/particle_beam.py
jank324: You mean the `dtype` PR? (I can't find the number right now)

jank324: #340

Hespe: Yes, #340. Maybe its not something we should address in this PR, but it seems fitting to have a similar interface to all converter and importer functions.

jank324: I opened a separate issue for this: #360

Hespe (approved):

CHANGELOG.md
```diff
@@ -38,6 +38,7 @@
 - Link to different ImpactX example in test docstring (see #341) (@ax3l)
 - Add link to the new Discord server (see #355) (@jank324)
 - Fix typo that said "quadrupole" in a dipole docstring (see #358) (@jank324)
+- Type annotations were updated to the post-PEP 585/604... style (see #359) (@jank324)
```
jank324, line 41:
```suggestion
- Type annotations were updated to the post-PEP 585/604... style (see #360) (@jank324)
```

PR #392, opened 2025-03-26
Plot beam mean and std in `plot_overview`

### LINKED ISSUE #391: Show beam mean and std in `plot_overview`, instead of reference particles
Showing the reference particles in `plot_overview` may be misleading, since these particle trajectories are not representative of the behavior of the whole beam. 

It would make sense to instead plot the mean and beam size.

RemiLehe:
This plots the beam mean position and beam size, instead of reference particles.

## Description

This replaces the function `plot_reference_particles` with a new function `plot_mean_and_std`:

Typical plot with the new function:
Screenshot 2025-03-26 at 6 27 10 AM

## Motivation and Context

Showing the reference particles in plot_overview may be misleading, since these particle trajectories are not representative of the behavior of the whole beam. The mean and std give a better sense of the overall behavior of the beam.

Closes #391

## Types of changes

## Checklist


cheetah/accelerator/segment.py
```diff
@@ -505,40 +503,60 @@ def plot_reference_particle_traces(
...
 
-        xs = [reference_beam.x for reference_beam in references]
-        broadcast_xs = torch.broadcast_tensors(*xs)
-        stacked_xs = torch.stack(broadcast_xs)
-        dimension_reordered_xs = stacked_xs.movedim(0, -1)  # Place vector dims first
-
-        ys = [reference_beam.y for reference_beam in references]
-        broadcast_ys = torch.broadcast_tensors(*ys)
-        stacked_ys = torch.stack(broadcast_ys)
-        dimension_reordered_ys = stacked_ys.movedim(0, -1)  # Place vector dims first
```
RemiLehe, line 521: I am not sure whether this rearrangement of dimensions is still needed, now that we plot the mean and std. @jank324 Any opinion?

jank324: I think it probably isn't. If I remember correctly it was only needed because the broadcasting placed the particle dimension ahead of the vector dimension. Now there is no particle dimension anymore.

jank324: Okay ... correction. The problem is that initially the dimension of samples along s comes first. We still need the vector dimension to come first. So I think the dimension rearrangement is still needed.

jank324:
For reference, here is the code I used now to generate the new README plot.

```python
import matplotlib.pyplot as plt
import torch
from ARESlatticeStage3v1_9 import cell

import cheetah


segment = cheetah.Segment.from_ocelot(cell).subcell("AREASOLA1", "ARMRBPMG2")

astra_beam = cheetah.ParticleBeam.from_astra(
    "../tests/resources/ACHIP_EA1_2021.1351.001"
)

segment.AREAMQZM1.k1 = torch.tensor(4.2)
segment.AREAMQZM2.k1 = torch.tensor(-10.0)
segment.AREAMQZM3.k1 = torch.tensor(4.2)

segment.AREAMQZM2.misalignment = torch.tensor([2e-4, 0.0])

segment.plot_overview(incoming=astra_beam)
plt.savefig("../images/readme_overview_plot.png", dpi=300)
plt.show()
```
readme_overview_plot

jank324 (approved):

PR #398, opened 2025-04-09
RPN rework

### LINKED ISSUE #387: Elegant parsing fails on property `body_focus_model = SRS` with unhandled error
`evaluate_expression()` in `fortran_namelist.py` fails with a `NameError` when parsing a property which could be a valid python variable name, as in "srs"

Just needs the except to catch `(SyntaxError, NameError)` to handle it properly. A better fix would be a proper RPN parser though instead of relying on `eval()`.

amylizzle:
## Description

Replaces the RPN parser with a proper implementation that can access context vars

## Motivation and Context

Closes #387 
Adds support for more complex nested expressions in RPN

## Types of changes

## Checklist


amylizzle:
Not an exact match for Elegant's RPN stuff [which is apparently huge](https://ops.aps.anl.gov/manuals/elegant_latest/elegantse12.html) but it's enough to parse most simple math and deref any vars

Hespe:
One of our lattices that can be imported with the current master fails to import on this branch. I have modified the test resource accordingly, hence the test failure.

From the looks of it, it was a happy accident that the line worked before (`eval` treating the `#` character as the beginning of a comment). I could not found the matching use of `#` in the elegant documentation, but would still like to preserve compatibility with our files.

amylizzle:
That's terrible, I love it. I'll get it fixed.

cheetah/converters/utils/fortran_namelist.py
```diff
@@ -137,18 +137,12 @@ def evaluate_expression(expression: str, context: dict) -> Any:
         # behaviour.
         expression = re.sub(r"abs\(", r"abs_func(", expression)
 
-        return (
-            eval(expression, context)
```
Hespe, line 141: Not quite sure, but I think this makes the above code for sanitizing the input for python irrelevant. Not sure if that is ok. I did not get any problems with a few Elegant files I tested, but maybe Bmad relies on it, @jank324 ?

cheetah/converters/utils/rpn.py
```diff
@@ -1,17 +1,49 @@
...
+                stack.append(a - b)
+            case "*":
+                b = stack.pop()
+                a = stack.pop()
+                stack.append(a * b)
+            case "/":
+                b = stack.pop()
+                a = stack.pop()
+                stack.append(a / b)
+            case "#":  # commment, ignore this and all following tokens
```
Hespe, line 30: That would miss instances where `#` is not surrounded by spaces, right? Might not be an issue as there are always spaces present in the files I checked.

amylizzle: I grabbed elegant's RPN parser application and tested against it.
`1 2 +#comment` fails to run but  `1 2 + #comment` works. 

I'll add handling for comments that don't have a space to be compliant.

cheetah/converters/utils/fortran_namelist.py
amylizzle: oh man I missed that - I'll have to add calling abs and exponentiation.

Hespe (commented):
Looks mostly good to me now. However, I would like someone to test if we somehow broke the `Bmad` parser before merging, since the expression replacement that got removed here seems to be meant for that.

I have the suspicion that at least some `Bmad` lattices are now broken, especially since there is no `abs` function in the new RPN parser, which apparently was needed somewhere in `Bmad`.

tests/test_rpn.py
```diff
@@ -7,8 +9,38 @@ def test_valid_rpn_expression():
...
 
-    assert rpn.is_valid_expression(expression)
+def test_complex_rpn_expression():
+    """
+    Test that a valid RPN expression with nesting is correctly recognised as a valid
+    RPN expression.
+    """
+    expression = "10 2 * 4 2 ^ + sqrt"  # sqrt(20 + 16) = 6
+    # try with empty context
+    assert rpn.try_eval_expression(expression, []) == 6
```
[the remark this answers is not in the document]
amylizzle: shouldn't matter in this case since this test should always be *exactly* 6.0

jank324:
@amylizzle, we added two tests that break the current implementation but work on `master`. They represent MWEs for cases that actually occur in the LCLS Bmad lattice files.

 - Use of `abs` as a function and variable name. It seems that in Fortran functions are not first class variables and therefore functions and variables can have the same name without causing conflict.
 - The use of parentheses in expressions, which if I followed this PR correctly, I believe you removed the parsing code for.

cheetah/converters/bmad.py
```diff
@@ -337,7 +337,7 @@ def convert_lattice_to_cheetah(
     ), "Merging lines should never produce more lines than there were before."
 
     # Parse the lattice file(s), i.e. basically execute them
-    context = parse_lines(merged_lines)
+    context = parse_lines(merged_lines, False)
```
Hespe, line 340:
```suggestion
    context = parse_lines(merged_lines, rpn_mode=False)
```

cheetah/converters/elegant.py
```diff
@@ -436,7 +436,7 @@ def convert_lattice_to_cheetah(
     stripped_lines = [line.strip(";") for line in merged_lines]
 
     # Parse the lattice file(s), i.e. basically execute them
-    context = parse_lines(stripped_lines)
+    context = parse_lines(stripped_lines, True)
```
Hespe, line 439:
```suggestion
    context = parse_lines(stripped_lines, rpn_mode=True)
```

cheetah/converters/utils/fortran_namelist.py
```diff
@@ -94,13 +94,14 @@ def merge_delimiter_continued_lines(
...
 
-def evaluate_expression(expression: str, context: dict) -> Any:
+def evaluate_expression(expression: str, context: dict, rpn_mode: bool) -> Any:
     """
     Evaluate an expression in the context of a dictionary of variables.
 
     :param expression: Expression to evaluate.
     :param context: Dictionary of variables to evaluate the expression in the context
         of.
+    :param rpn_mode: whether expressions should be treated as RPN
```
Hespe, line 104:
```suggestion
    :param rpn_mode: Whether expressions should be treated as reverse Polish notation.
```

cheetah/converters/utils/fortran_namelist.py
```diff
@@ -162,13 +177,14 @@ def resolve_object_name_wildcard(wildcard_pattern: str, context: dict) -> list:
...
 
-def assign_property(line: str, context: dict) -> dict:
+def assign_property(line: str, context: dict, rpn_mode: bool) -> dict:
     """
     Assign a property of an element to the context.
 
     :param line: Line of a property assignment to be parsed.
     :param context: Dictionary of variables to assign the property to and from which to
         read variables.
+    :param rpn_mode: whether expressions should be treated as RPN
```
Hespe, line 187:
```suggestion
    :param rpn_mode: Whether expressions should be treated as reverse Polish notation.
```

cheetah/converters/utils/fortran_namelist.py
```diff
@@ -208,18 +225,19 @@ def assign_variable(line: str, context: dict) -> dict:
...
 
-def define_element(line: str, context: dict) -> dict:
+def define_element(line: str, context: dict, rpn_mode: bool) -> dict:
     """
     Define an element in the context.
 
     :param line: Line of an element definition to be parsed.
     :param context: Dictionary of variables to define the element in and from which to
         read variables.
+    :param rpn_mode: whether expressions should be treated as RPN
```
Hespe, line 240:
```suggestion
    :param rpn_mode: Whether expressions should be treated as reverse Polish notation.
```

cheetah/converters/utils/fortran_namelist.py
```diff
@@ -347,12 +366,14 @@ def parse_use_line(line: str, context: dict) -> dict:
...
-def parse_lines(lines: str) -> dict:
+def parse_lines(lines: str, rpn_mode: bool) -> dict:
     """
-    Parse a list of lines from a Bmad lattice file. They should be cleaned and merged
-    before being passed to this function.
+    Parse a list of lines from a Bmad or Elegant lattice file. They should be cleaned
+    and merged before being passed to this function.
 
     :param lines: List of lines to parse.
+    :param rpn_mode: Whether expressions should be parsed as RPN.
```
Hespe, line 375:
```suggestion
    :param rpn_mode: Whether expressions should be parsed as reverse Polish notation.
```

Hespe (approved):

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 ### 🚀 Features
 
 - Implement `split` method for the `Solenoid` element (see #380) (@cr-xu)
+- Implement a more robust RPN parser, fixing a bug where short strings in an Elegant variable definition would cause parsing to fail (see #387) (@amylizzle)
```
jank324, line 12:
```suggestion
- Implement a more robust RPN parser, fixing a bug where short strings in an Elegant variable definition would cause parsing to fail. (see #387) (@amylizzle)
```

jank324 (approved):

cheetah/converters/utils/fortran_namelist.py
```diff
@@ -126,11 +125,13 @@ def evaluate_expression(expression: str, context: dict, is_rpn: bool = False) ->
...
     # Evaluate as a mathematical expression
+    # TODO I would prefer if it tried to do normal parsing first and RPN second.
+    #   Implement when normal parsing was made safe.
     try:
-        if is_rpn:
-            return rpn.evaluate_expression(expression, context)
-        else:
-            # TODO: This is currently an unsafe eval and should be replaced with a safer
+        return rpn.evaluate_expression(expression, context)
+    except SyntaxError:
```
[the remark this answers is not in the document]
jank324: What do you suggest instead? I didn't like passing the `rpn_mode` through the entire stack, because I think Bmad and Elegant allow you to use either. So I would like it to detect automatically on a per-expression basis.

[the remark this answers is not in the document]
amylizzle: ah, I didn't realise they could each do either. 
Yeah, I guess in that case this is the best approach. We'll just have to write an infix parser asap to prevent odd behaviour on a  bad RPN parse.

PR #415, opened 2025-04-24
Warn on not-understood properties

amylizzle:
## Description

Swaps misunderstood properties from an assert to a print, to match behaviour of not understood elements.

## Motivation and Context
There are several properties in the CLARA/FEBE lattice that are not understood and don't seem that important to Cheetah

## Types of changes

## Checklist

### DIFF
--- a/cheetah/converters/utils/fortran_namelist.py
+++ b/cheetah/converters/utils/fortran_namelist.py
@@ -423,8 +423,11 @@ def validate_understood_properties(understood: list[str], properties: dict) -> N
     :return: None
     """
     for property in properties:
-        assert any([re.fullmatch(pattern, property) for pattern in understood]), (
-            f"Property {property} with value {properties[property]} for element type"
-            f" {properties['element_type']} is currently not understood. Other values"
-            f" in properties are {properties.keys()}."  # noqa: B038
-        )
+        if any([re.fullmatch(pattern, property) for pattern in understood]):
+            continue
+        else:
+            print(
+                f"WARNING: Property {property} with value {properties[property]}"
+                f" for element type {properties['element_type']} is currently"
+                f" not understood."
+            )


cheetah/converters/utils/fortran_namelist.py
```diff
@@ -439,8 +439,11 @@ def validate_understood_properties(understood: list[str], properties: dict) -> N
...
-            f" in properties are {properties.keys()}."  # noqa: B038
-        )
+        if any([re.fullmatch(pattern, property) for pattern in understood]):
+            continue
+        else:
+            print(
+                f"WARNING: Property {property} with value {properties[property]}"
+                f" for element type {properties['element_type']} is currently"
+                f" not understood."
+            )
```
jank324, line 449: This should listen to the `warnings` option.

jank324 (changes requested):
See comment.

jank324:
@Hespe and I realised that there is one more thing that this PR needs to do. Right now, there are a bunch of properties that we encountered when implementing the converter and then silently ignored. The error is only thrown for properties that we never encountered before, meaning that now a warning would only be issued for properties we've never seen before.

Instead, a warning (specifically a `PhysicsWarning`) should be issued for all properties that might result in Cheetah producing different physics. Properties that don't do that can still be silently ignored.

CHANGELOG.md
```diff
@@ -35,6 +35,7 @@
 - Rotation matrices are no longer computed twice for forward and backward phase space rotations (see #452) (@Hespe)
 
 ### 🐛 Bug fixes
+- Unrecognised element properties now print a warning instead of exiting with an assertation error (see #425) (@amylizzle)
```
jank324, line 38: We shouldn't forget to move this up to the new release.

Hespe (approved):

jank324:
@Hespe can you have a look if the tolerance increase seems "responsible"? Then this PR can be merged.

PR #418, opened 2025-04-25
Fixture for creating MWE cheetah elements in tests

### LINKED ISSUE #416: Automatically execute tests for all cheetah elements
There are some tests that should be ran for all cheetah elements, i.e. tests for `clone` and `dtype`. Currently these tests are parametrized by a manually curated lists of elements, which is prone to get outdated if new elements are added (see #406).

Instead, we should implement a test fixture that automatically runs a test with all available cheetah elements, including necessary default arguments. This could take inspiration of the tests in `test_elements`.

Hespe:
## Description

Some tests in the cheetah testsuite should be executed for all subclasses of `Element`. To make this convenient, we implement a fixture that automatically generates mimimal working examples of all elements and runs the test with them.

## Motivation and Context

  - Closes #416 

## Types of changes

## Checklist

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -3,7 +3,27 @@ import random
 import pytest
 import torch
 
+import cheetah
 from cheetah.utils import is_mps_available_and_functional
+
+ELEMENT_CLASS_DEFAULT_ARGS = {
+    cheetah.Aperture: {},
+    cheetah.BPM: {},
+    cheetah.Cavity: {"length": torch.tensor(1.0)},
+    cheetah.CustomTransferMap: {"predefined_transfer_map": torch.eye(7)},
+    cheetah.Drift: {"length": torch.tensor(1.0)},
+    cheetah.Dipole: {"length": torch.tensor(1.0)},
+    cheetah.HorizontalCorrector: {"length": torch.tensor(1.0)},
+    cheetah.Marker: {},
+    cheetah.Quadrupole: {"length": torch.tensor(1.0)},
+    cheetah.Segment: {"elements": [cheetah.Drift(length=torch.tensor(1.0))]},
+    cheetah.Screen: {},
+    cheetah.Solenoid: {"length": torch.tensor(1.0)},
+    cheetah.SpaceChargeKick: {"effect_length": torch.tensor(1.0)},
+    cheetah.TransverseDeflectingCavity: {"length": torch.tensor(1.0)},
+    cheetah.Undulator: {"length": torch.tensor(1.0)},
+    cheetah.VerticalCorrector: {"length": torch.tensor(1.0)},
+}
 
 
 def pytest_addoption(parser):
@@ -48,3 +68,19 @@ def default_torch_dtype(request):
     yield tmp_dtype_for_test
 
     torch.set_default_dtype(previous_dtype)
+
+
+@pytest.fixture(params=cheetah.Element.__subclasses__())
+def mwe_cheetah_element(request):
+    """
+    Run a test with a minimum working example of every cheetah Element. A default
+    value is used for elements with mandatory arguments.
+    """
+    ElementClass = request.param
+
+    if ElementClass in ELEMENT_CLASS_DEFAULT_ARGS:
+        return ElementClass(**ELEMENT_CLASS_DEFAULT_ARGS[ElementClass])
+    else:
+        raise NotImplementedError(
+            f"No default arguments for element class '{ElementClass}'"
+        )
--- a/tests/test_clone.py
+++ b/tests/test_clone.py
@@ -4,30 +4,14 @@ import torch
 import cheetah
 
 
-@pytest.mark.parametrize(
-    "ElementClass",
-    [
-        cheetah.Cavity,
-        cheetah.Dipole,
-        cheetah.Drift,
-        cheetah.HorizontalCorrector,
-        cheetah.Quadrupole,
-        cheetah.RBend,
-        cheetah.Solenoid,
-        cheetah.TransverseDeflectingCavity,
-        cheetah.Undulator,
-        cheetah.VerticalCorrector,
-    ],
-)
-def test_element_buffer_contents_and_location(ElementClass):
+def test_element_buffer_contents_and_location(mwe_cheetah_element):
     """
     Test that the buffers of cloned elements have the same content while not sharing the
     same memory location.
     """
-    element = ElementClass(length=torch.tensor(1.0))
-    clone = element.clone()
+    clone = mwe_cheetah_element.clone()
 
-    for buffer, buffer_clone in zip(element.buffers(), clone.buffers()):
+    for buffer, buffer_clone in zip(mwe_cheetah_element.buffers(), clone.buffers()):
         assert torch.allclose(buffer, buffer_clone)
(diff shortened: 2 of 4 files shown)


jank324 (changes requested):
Try to rewrite as marker

Hespe:
> Try to rewrite as marker

There is now a mandatory `test_all_elements` marker you have to put before tests using the `mwe_cheetah_element` fixture for it to run. Also the `skip_element` marker is replaced by the optional keyword argument `except_for`.

The marker/fixture combination is flexible enough to also allow for additional arguments later, e.g., `pass_argument`. This would allow to make even more tests a catch all version to prevent us from missing elements in the tests again. Maybe even a whitelist argument like `only` as an alternative to the blacklist argument `except_for`, but I am not sure about that one.

tests/conftest.py
```diff
@@ -3,16 +3,69 @@
...
+        argument_name = "mwe_element"
+        element_classes = cheetah.Element.__subclasses__()
+
+        # Handle keyword arguments
+        for keyword, value in element_marker.kwargs.items():
+            match (keyword):
+                case "arg_name":
+                    argument_name = value
+                case "except_for":
+                    for clazz in value:
```
jank324, line 53: `clazz`? WTF! 😂

tests/test_clone.py
```diff
@@ -4,32 +4,16 @@
...
     same memory location.
     """
-    element = ElementClass(length=torch.tensor(1.0))
-    clone = element.clone()
+    clone = mwe_element.clone()
 
-    for buffer, buffer_clone in zip(element.buffers(), clone.buffers()):
-        assert torch.allclose(buffer, buffer_clone)
+    for buffer, buffer_clone in zip(mwe_element.buffers(), clone.buffers()):
+        assert torch.allclose(buffer, buffer_clone, equal_nan=True)
```
jank324, line 16: Why was the `equal_nan` needed? That's sus ...

tests/test_device_dtype.py
```diff
@@ -22,127 +25,135 @@
...
+    """Test that elements can be successfully moved to a different device."""
 
-    # Test that by default the quadrupole is on the CPU
-    assert quad.length.device.type == "cpu"
-    assert quad.k1.device.type == "cpu"
-    assert quad.misalignment.device.type == "cpu"
-    assert quad.tilt.device.type == "cpu"
+    # Test that by default the element is on the CPU
+    for buffer in mwe_element.buffers():
+        assert buffer.dtype == "cpu"
```
jank324, line 33: I think I would keep this test as it was? Testing the buffers is just testing if `torch` works as expected, which isn't the job of Cheetah.

tests/conftest.py
```diff
@@ -36,34 +36,52 @@ def pytest_addoption(parser) -> None:
 
 def pytest_generate_tests(metafunc: pytest.Metafunc) -> None:
     """
-    Add a parametrize like marker to the test suite that automatically initializes
-    minimal working examples of cheetah elements.
+    Add a `pytest.mark.parametrize`-like marker that runs a test with an MWE of each
```
jank324, line 39: Is there a system to the order in this file? I feel like all the stuff for the marker feature in this PR is all over the place when it should be together. Or am I missing something?

jank324:
Also, my refactoring broke the tests and I don't understand why. Especially the missing argument `self` error confuses me a lot. 🤔

tests/conftest.py
Hespe: I see you don't know you way around in the [Java world](https://stackoverflow.com/questions/2529974/why-do-java-programmers-like-to-name-a-variable-clazz)

tests/test_clone.py
Hespe: That was for handling the `BPM` properly. Its `reading` property is initialized with `NaN` and would thus fail the test even if the clone was successful. Another solution would of course be to track once before cloning, but I deemed that to involved for such a simple test.

tests/test_device_dtype.py
Hespe: Fine with me. In that case we are restricted to test only `length` though, since that is the only property shared by all elements.

tests/conftest.py
Hespe: I kept the pytest hooks at the top, i.e. `pytest_addoption`, `pytest_generate_tests`, and `pytest_report_header`. And below that it's the custom fixtures.

Maybe sorting stuff by its purpose would be more elegant. Currently there are three things: PRNG initialization, default `dtype` initialization, and MWE element creation.

tests/test_device_dtype.py
Hespe: All calls to `buffers` have been removed.

cheetah/accelerator/space_charge_kick.py
```diff
@@ -46,35 +44,50 @@ class SpaceChargeKick(Element):
...
-        num_grid_points_y: int = 32,
-        num_grid_points_tau: int = 32,
-        grid_extend_x: torch.Tensor = 3,  # TODO: Simplify these to a single tensor?
-        grid_extend_y: torch.Tensor = 3,
-        grid_extend_tau: torch.Tensor = 3,
+        grid_shape: tuple[int, int, int] = (32, 32, 32),
+        # TODO: Simplify these to a single tensor?
+        grid_extend_x: torch.Tensor | None = None,
+        grid_extend_y: torch.Tensor | None = None,
+        grid_extend_tau: torch.Tensor | None = None,
```
jank324, line 51: Should probably be `extent`, right?

tests/test_clone.py
```diff
@@ -39,12 +26,15 @@ def test_beam_buffer_contents_and_location(BeamClass):
...
     same memory location.
     """
-    beam = BeamClass.from_parameters(species=cheetah.Species("proton"))
+    beam = BeamClass.from_parameters(species=cheetah.Species("positron"))
     clone = beam.clone()
 
-    for buffer, buffer_clone in zip(beam.buffers(), clone.buffers()):
-        assert torch.allclose(buffer, buffer_clone)
-        assert not buffer.data_ptr() == buffer_clone.data_ptr()
+    for attribute in beam.UNVECTORIZED_NUM_ATTR_DIMS.keys():
```
jank324, line 32: You call it `feature` for `Element` and `attribute` for `Beam`?

cheetah/accelerator/space_charge_kick.py
```diff
@@ -31,50 +31,63 @@ class SpaceChargeKick(Element):
...
         self,
         effect_length: torch.Tensor,
-        num_grid_points_x: int = 32,  # TODO: Simplify these to a single tuple?
-        num_grid_points_y: int = 32,
-        num_grid_points_tau: int = 32,
-        grid_extend_x: torch.Tensor = 3,  # TODO: Simplify these to a single tensor?
-        grid_extend_y: torch.Tensor = 3,
-        grid_extend_tau: torch.Tensor = 3,
+        grid_shape: tuple[int, int, int] = (32, 32, 32),
+        # TODO: Simplify these to a single tensor?
```
jank324, line 48: Kann das weg?

jank324: Ne, kann nicht.

Issue #395, opened 2025-03-30
Screen image orientation is incorrect


roussel-ryan:
Using the cheetah screen element with a `ParticleBeam` and mode "histogram" or "kde" currently outputs the image in an incorrect orientation with respect to the `pixel_bin_centers` property. This can be determined from the following:

I created a beam with a strong x-y correlation as shown below, such that the beam is much larger in x than in y.
![Image](https://github.com/user-attachments/assets/d8e5a033-f9d6-45eb-a773-f4df18c03d0e)

I tracked the beam through the following segment
```python
from cheetah.particles import ParticleBeam
from cheetah.accelerator import Segment, Screen
import torch

seg = Segment([
    Screen(resolution=(500,500), pixel_size=torch.ones(2)*50e-6, is_active=True, method="histogram", kde_bandwidth=torch.tensor(50e-6)),
])

beam = ParticleBeam.from_twiss(
    beta_x=torch.tensor(1.0), beta_y=torch.tensor(10.0), alpha_x=torch.tensor(0.0), alpha_y=torch.tensor(0.0),
    emittance_x=torch.tensor(1e-9), emittance_y=torch.tensor(1e-7),
    num_particles=10000, total_charge=torch.tensor(1e-9),
)
beam.particles[:, 0] += beam.particles[:,2] * 5.0
seg.track(beam);
```
and then used matplotlibs pcolor method to plot the image using the pixel bin centers and the screen reading
```python
import matplotlib.pyplot as plt
plt.pcolor(*seg.elements[0].pixel_bin_centers, seg.elements[0].reading)
plt.ylabel("y (m)")
plt.xlabel("x (m)")
```
Which produced the following plot

![Image](https://github.com/user-attachments/assets/0141463a-e3f1-4571-ac6a-cf3bb0efa623)

Clearly the particle distribution and plot disagree. The correct transformation for at least the kde measurement is the following
```python
image = kde_histogram_2d(
    x1=broadcasted_x,
    x2=broadcasted_y,
    bins1=self.pixel_bin_centers[0].to(read_beam.x),
    bins2=self.pixel_bin_centers[1].to(read_beam.x),
    bandwidth=self.kde_bandwidth,
    weights=broadcasted_weights,
)
# Change the x, y positions
image = torch.transpose(image, -2, -1)
```
not with the additional flip operation.

roussel-ryan:
@jp-ga @lboult

PR #428, opened 2025-04-28
fix `Screen` device issue

### LINKED ISSUE #427: `Screen` mixes gpu and cpu devices
After PR #395 , `Screen` element mixes gpu and cpu devices.

jp-ga:
## Description

After PR #395 , `Screen` element mixes gpu and cpu devices.

## Motivation and Context

Fixes #427 

## Types of changes

## Checklist

### DIFF
--- a/cheetah/accelerator/screen.py
+++ b/cheetah/accelerator/screen.py
@@ -185,10 +185,10 @@ class Screen(Element):
 
                 copy_of_incoming.particles[..., 0] -= self.misalignment[
                     ..., 0
-                ].unsqueeze(-1)
+                ].unsqueeze(-1).to(copy_of_incoming.particles)
                 copy_of_incoming.particles[..., 2] -= self.misalignment[
                     ..., 1
-                ].unsqueeze(-1)
+                ].unsqueeze(-1).to(copy_of_incoming.particles)
 
             self.set_read_beam(copy_of_incoming)
 
@@ -296,8 +296,8 @@ class Screen(Element):
                 image = kde_histogram_2d(
                     x1=broadcasted_x,
                     x2=broadcasted_y,
-                    bins1=self.pixel_bin_centers[0],
-                    bins2=self.pixel_bin_centers[1],
+                    bins1=self.pixel_bin_centers[0].to(read_beam.x),
+                    bins2=self.pixel_bin_centers[1].to(read_beam.x),
                     bandwidth=self.kde_bandwidth,
                     weights=broadcasted_weights,
                 )
--- a/cheetah/accelerator/screen.py
+++ b/cheetah/accelerator/screen.py
@@ -183,12 +183,16 @@ class Screen(Element):
                 )
                 copy_of_incoming.particles = broadcasted_particles.clone()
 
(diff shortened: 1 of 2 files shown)


cheetah/accelerator/screen.py
```diff
@@ -183,12 +183,16 @@ def track(self, incoming: Beam) -> Beam:
...
-                copy_of_incoming.particles[..., 0] -= self.misalignment[
-                    ..., 0
-                ].unsqueeze(-1)
-                copy_of_incoming.particles[..., 2] -= self.misalignment[
-                    ..., 1
-                ].unsqueeze(-1)
+                copy_of_incoming.particles[..., 0] -= (
+                    self.misalignment[..., 0]
+                    .unsqueeze(-1)
+                    .to(copy_of_incoming.particles)
```
jank324, line 189: @jp-ga, could you add a test that would catch the issue?

The `to()` here shouldn't be needed if `copy_of_incoming.particles` and `self.misalignment` are on the correct device. So I have a feeling, that this fix only addresses the symptom but not the root cause.

jank324 (changes requested):

jp-ga:
As reference, this is the error I get when running the main branch version: 

```
File ~/repos/cheetah/cheetah/accelerator/screen.py:186, in Screen.track(self, incoming)
    180     broadcasted_particles, _ = torch.broadcast_tensors(
    181         copy_of_incoming.particles,
    182         self.misalignment[..., 0].unsqueeze(-1).unsqueeze(-1),
    183     )
    184     copy_of_incoming.particles = broadcasted_particles.clone()
--> 186     copy_of_incoming.particles[..., 0] -= (
    187         self.misalignment[..., 0]
    188         .unsqueeze(-1)
    189     )
    190     copy_of_incoming.particles[..., 2] -= (
    191         self.misalignment[..., 1]
    192         .unsqueeze(-1)
    193     )
    195 self.set_read_beam(copy_of_incoming)

RuntimeError: Expected all tensors to be on the same device, but found at least two devices, mps:0 and cpu!
```

cheetah/accelerator/screen.py
jp-ga: ok, will take a look into it

jp-ga: What happens is that the screen is created in CPU, so misalignment is in CPU. Then, when trying to track an incoming beam with particles in GPU, the screen misalignment doesn't change automatically. I can add a line of code so that the screen misalignment changes the device if it is tracking particles that are in GPU.

jank324: It is the user's responsibility to make sure the beam and element are on the same device. This would be similar behaviour to PyTorch itself, where you are also responsible for making sure that the input to a e.g. neural network and the neural network model itself are both on the same device.

jank324: Now I'm wondering: Is this still a bug at all, or was it just the result of you assuming that Cheetah will handle this automatically?

jp-ga: the latter. I will close the PR

PR #432, opened 2025-04-29
Method for reducing number of particles in a beam

jank324:
## Description

Implements a method that allows the user to reduce the number of particles in a `ParticleBeam` for improved computational efficiency and memory use.

## Motivation and Context

It's easy to run out of memory when doing vectorised computations with large `ParticleBeam`s.

## Types of changes

## Checklist


CHANGELOG.md
```diff
@@ -26,6 +26,7 @@
 - Implement `split` method for the `Solenoid` element (see #380) (@cr-xu)
 - Implement a more robust RPN parser, fixing a bug where short strings in an Elegant variable definition would cause parsing to fail. (see #387, #417) (@amylizzle, @Hespe, @jank324)
 - Add a `Sextupole` element (see #406) (@jank324, @Hespe)
+- Add a method to `ParticleBeam` that lets you subsample a particle beam with fewer particles and the same distribution (see #432) (@jank324)
```
Hespe, line 29: This should be in 0.7.3

cheetah/particles/particle_beam.py
```diff
@@ -1033,6 +1033,56 @@ def linspaced(self, num_particles: int) -> "ParticleBeam":
             dtype=self.particles.dtype,
         )
 
+    def randomly_subsampled(
+        self,
+        num_particles: int,
+        adjust_particle_charges: bool = True,
+        random_state: torch.Generator | None = None,
```
Hespe, line 1040: We have not passed the RNG in the same way anywhere else. Is this something we want to be doing in the future? Or should we keep the API consistent and just use the default torch RNG?

cheetah/particles/particle_beam.py
```diff
@@ -1033,6 +1033,56 @@ def linspaced(self, num_particles: int) -> "ParticleBeam":
...
+    ) -> "ParticleBeam":
+        """
+        Create a new beam with the same parameters as this beam, but with
+        `num_particles` particles randomly sampled from the original beam.
+
+        :param num_particles: Number of particles to create.
```
Hespe, line 1046:
```suggestion
        :param num_particles: Number of particles to sample.
```

cheetah/particles/particle_beam.py
```diff
@@ -1033,6 +1033,56 @@ def linspaced(self, num_particles: int) -> "ParticleBeam":
...
+        :param adjust_particle_charges: If True, the particle charges are adjusted
+            to match the total charge of the new beam.
```
Hespe, line 1048:
```suggestion
            to match the total charge of the old beam.
```

Maybe I misunderstood, but I guess the idea is to create new particles with larger charge such that the total charge of the old beam is preserved with fewer particles?

cheetah/particles/particle_beam.py
```diff
@@ -1033,6 +1033,56 @@ def linspaced(self, num_particles: int) -> "ParticleBeam":
...
+            "Number of particles to sample must be less than or equal to the number of "
+            "particles in the original beam."
+        )
+
+        if random_state is None:
+            random_state = torch.Generator(device=self.particles.device)
+
+        random_indices = torch.randint(
+            0, self.particles.shape[-2], (num_particles,), generator=random_state
+        )
```
Hespe, line 1063: This might sample the same particle multiple times if I am not mistaken. Should that be this way? I would have expected to get the same particles if `num_particles` matches the number of particles of the original beam, not a subset with some repeating copies.

I think the other result could be achieved by using `torch.randperm` and cutting off the result after `num_particles` entries.

cheetah/particles/particle_beam.py
```diff
@@ -1033,6 +1033,56 @@ def linspaced(self, num_particles: int) -> "ParticleBeam":
...
+            energy=self.energy,
+            particle_charges=subsampled_particle_charges,
+            survival_probabilities=subsampled_survival_probabilities,
+            species=self.species,
+            device=self.particles.device,
+            dtype=self.particles.dtype,
+        )
+
+        if adjust_particle_charges:
+            subsampled_beam.particle_charges *= (
```
Hespe, line 1080: I had not thought about this before, but how does this compare to the charge of `subsampled_beam.species`? We are essentially now dealing with some kind of macroparticle, right, so should this be reflected in the `Species`?

Hespe (changes requested):
One more general remark about the test. We are currently just comparing the first to stochastic moments, so I guess this test could potentially also be passed by fitting a Gaussian beam.

cheetah/particles/particle_beam.py
jank324: I'm not sure. I quite like the idea of explicit randomness since I saw it in JAX. Here is was a proposal by GitHub Copilot and I kind of liked it.

Maybe we treat this as a trial location, and if we like it, we transfer it to other places across Cheetah as well?

cheetah/particles/particle_beam.py
jank324: I guess it shouldn't be reflected. The species charge is a physical charge of the real particle, that we are representing. The charge of a macroparticle on the other hand is not really physical. So it should be possible to figure out how many physical particles are represented by a macroparticle through `particle_charges[particle_idx] / species.charge_coulomb` ... at least I think so.

cheetah/particles/particle_beam.py
Hespe: Its not quite the explicit randomness of JAX, but I agree, it gives the option to go in a similar direction. And since its optional it is of course easy to ignore and recover the behaviour of other methods.

jank324:
> One more general remark about the test. We are currently just comparing the first to stochastic moments, so I guess this test could potentially also be passed by fitting a Gaussian beam.

Do you think we should improve the comparison?

Hespe:
> > One more general remark about the test. We are currently just comparing the first to stochastic moments, so I guess this test could potentially also be passed by fitting a Gaussian beam.
> 
> Do you think we should improve the comparison?

I have a suspicion that the current test could be passed by a roundtrip conversion using `ParameterBeam`, e.g. `ParticleBeam` -> `ParameterBeam` -> `ParticleBeam`, which would always result in a Gaussian beam. The new method should be much better for highly skewed distributions.

cheetah/particles/particle_beam.py
jank324: Fixed!

jank324:
@Hespe how about now?

Hespe (approved):
I still believe that the test could be passed by fitting a Gaussian beam but don't know of any better alternative. So its probably good enough for our purpose.

PR #446, opened 2025-05-21
Correct beam duration in space-charge test (only affects non-relativistic case)

RemiLehe:
The way the space-charge tests where written was incorrect for the non-relativistic case.

## Description

The space-charge tests represents a situation where the beam has the same size in `x`, `y` and `z`, in the beam rest frame. (See Free Expansion of a Cold Uniform Bunch in https://accelconf.web.cern.ch/hb2023/papers/thbp44.pdf.) But since `cheetah` works with the beam duration, we need to correctly calculate the corresponding beam duration. 

There was a missing `beta` factor in the previous formula. This only matters for non-relativistic beams.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 ### 🐛 Bug fixes
 
 - Fix issue where `Dipole` with `tracking_method="bmadx"` and `angle=0.0` would output `NaN` values as a result of a division by zero (see #434) (@jank324)
+- Fix issue in CI space-charge tests (incorrect beam duration in non-relativistic case) (see #446) (@RemiLehe)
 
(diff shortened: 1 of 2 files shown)


RemiLehe:
@cr-xu @jank324 
The test failures seem unrelated to my PR (they come from ` tests\test_particle_beam.py`, which I did not modify); any idea why?

cr-xu (changes requested):
@RemiLehe  Can you add a line in the Changelog. Apart from that it looks good to me and I can merge the PR.

RemiLehe:
@cr-xu Sounds good ; done.

cr-xu (approved):

PR #450, opened 2025-05-27
Use builtin python logging module

### LINKED ISSUE #449: Use python `logging` module instead of `print`
In a few places, Cheetah is using `print` statements to issue warnings to the user, e.g. the lattice converters. Unfortunately, this in not integrated with the standard python `logging` module such that these warnings fail to appear in log files.

If instead, we were using the `logging` module on its `warning` level, the user could decide whether to save Cheetah warnings with their other logs or they would still be issued to `stdout` by default.

Hespe:
## Description

Use the builtin python `warnings` module instead of the standard `print` function. This makes the warnings in Cheetah more flexible, e.g., downstream users can silence warnings globally and much more

## Motivation and Context

  - Closes #449 

## Types of changes

## Checklist

### DIFF
--- a/cheetah/converters/bmad.py
+++ b/cheetah/converters/bmad.py
@@ -17,7 +17,7 @@ from cheetah.utils import PhysicsWarning
 def convert_element(
     name: str,
     context: dict,
-    sanitze_name: bool = False,
+    sanitize_name: bool = False,
     device: torch.device | None = None,
     dtype: torch.dtype | None = None,
 ) -> "cheetah.Element":
@@ -25,7 +25,7 @@ def convert_element(
 
     :param name: Name of the (top-level) element to convert.
     :param context: Context dictionary parsed from Bmad lattice file(s).
-    :param sanitze_name: Whether to sanitise the name to be a valid Python variable
+    :param sanitize_name: Whether to sanitise the name to be a valid Python variable
         name.
     :param device: Device to put the element on. If `None`, the current default device
         of PyTorch is used.
@@ -43,11 +43,11 @@ def convert_element(
     if isinstance(bmad_parsed, list):
         return cheetah.Segment(
             elements=[
-                convert_element(element_name, context, sanitze_name, device, dtype)
+                convert_element(element_name, context, sanitize_name, device, dtype)
                 for element_name in bmad_parsed
             ],
             name=name,
-            sanitize_name=sanitze_name,
+            sanitize_name=sanitize_name,
         )
     elif isinstance(bmad_parsed, dict) and "element_type" in bmad_parsed:
         if bmad_parsed["element_type"] == "marker":
@@ -62,7 +62,7 @@ def convert_element(
                 ],
                 bmad_parsed,
             )
-            return cheetah.Marker(name=name, sanitize_name=sanitze_name)
+            return cheetah.Marker(name=name, sanitize_name=sanitize_name)
         elif bmad_parsed["element_type"] == "monitor":
             validate_understood_properties(
                 ["element_type", "alias", "type", "l"], bmad_parsed
@@ -71,10 +71,10 @@ def convert_element(
                 return cheetah.Drift(
                     length=torch.tensor(bmad_parsed["l"], **factory_kwargs),
                     name=name,
-                    sanitize_name=sanitze_name,
+                    sanitize_name=sanitize_name,
                 )
             else:
-                return cheetah.Marker(name=name, sanitize_name=sanitze_name)
+                return cheetah.Marker(name=name, sanitize_name=sanitize_name)
         elif bmad_parsed["element_type"] == "instrument":
             validate_understood_properties(
                 ["element_type", "alias", "type", "l"], bmad_parsed
@@ -83,10 +83,10 @@ def convert_element(
                 return cheetah.Drift(
                     length=torch.tensor(bmad_parsed["l"], **factory_kwargs),
                     name=name,
-                    sanitize_name=sanitze_name,
+                    sanitize_name=sanitize_name,
                 )
             else:
-                return cheetah.Marker(name=name, sanitize_name=sanitze_name)
+                return cheetah.Marker(name=name, sanitize_name=sanitize_name)
         elif bmad_parsed["element_type"] == "pipe":
             validate_understood_properties(
                 ["element_type", "alias", "type", "l", "descrip"], bmad_parsed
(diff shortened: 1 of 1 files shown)


Hespe:
For some of these, a [`Warning`](https://docs.python.org/3/library/warnings.html#module-warnings) might actually more suitable.

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@
 - Bmad is no longer actively run in the test workflows, and comparisons to Bmad are now done on static pre-computed results from Bmad. This also removes the use of Anaconda in the test workflow. (see #429, #431) (@jank324)
 - The PyTorch pin to `<=2.6` was removed, as the issue with `abort trap: 6` was caused by Bmad is no longer actively used in the test workflow (see #429, #431) (@jank324)
 - There was a temporary pin `snowballstemmer<3.0` for the docs build because of an issue with the latest release. It has since been unpinned again because the release was yanked. Refer to https://github.com/sphinx-doc/sphinx/issues/13533 and https://github.com/snowballstem/snowball/issues/229. (see #436, #438) (@jank324)
+- Warning messages are now produced using the `logging` module of Python. Downstream users can therefore now filter and redirect the issues log entries. (see #450) (@Hespe)
```
jank324, line 26: I think this is a feature.

cheetah/converters/ocelot.py
```diff
@@ -1,7 +1,11 @@
+import logging
+
 import torch
 
 import cheetah
 
+log = logging.getLogger(__name__)
```
jank324, line 7: Is `log` the standard name given here?

cheetah/converters/utils/fortran_namelist.py
```diff
@@ -150,13 +153,13 @@ def evaluate_expression(expression: str, context: dict, warnings: bool = True) -
...
-                    f"WARNING: Could not evaluate expression {expression}. It will now "
-                    "be treated as a string. This may lead to unexpected behaviour."
+                log.warning(
+                    f"Could not evaluate expression '{expression}'. It will now be"
+                    " treated as a string. This may lead to unexpected behaviour."
                 )
             return expression
         except Exception as e:
-            print(expression)
+            log.exception(f"Failed to evaluate expression '{expression}'")
```
jank324, line 162: Actually, why are we printing here at all? Shouldn't the exception just stop execution and then be shown?

jank324 (changes requested):
What are the arguments `logging` vs `warnings`?

Hespe:
> What are the arguments `logging` vs `warnings`?

From https://docs.python.org/3/howto/logging.html:

> [warnings.warn()](https://docs.python.org/3/library/warnings.html#warnings.warn) in library code if the issue is avoidable and the client application should be modified to eliminate the warning
>
> A logger’s [warning()](https://docs.python.org/3/library/logging.html#logging.Logger.warning) method if there is nothing the client application can do about the situation, but the event should still be noted

jank324:
Think about adding `sanitze_names` argument to importers.

cheetah/accelerator/screen.py
```diff
@@ -33,6 +33,9 @@ class Screen(Element):
         distribution. If `False` the screen is inactive and will not record the beam's
         distribution.
     :param name: Unique identifier of the element.
+    :param sanitize_name: Whether to sanitisze the name to be a valid Python
```
jank324, line 36:
```suggestion
    :param sanitize_name: Whether to sanitise the name to be a valid Python
```

cheetah/accelerator/solenoid.py
```diff
@@ -26,6 +26,9 @@ class Solenoid(Element):
     :param misalignment: Misalignment vector of the solenoid magnet in x- and
         y-directions.
     :param name: Unique identifier of the element.
+    :param sanitize_name: Whether to sanitisze the name to be a valid Python
```
jank324, line 29:
```suggestion
    :param sanitize_name: Whether to sanitise the name to be a valid Python
```

cheetah/accelerator/segment.py
```diff
@@ -292,8 +300,9 @@ def to_lattice_json(
     def from_ocelot(
         cls,
         cell,
+        sanitize_element_names: bool = False,
         name: str | None = None,
-        warnings: bool = True,
+        sanitize_name: bool = False,
```
Hespe, line 305: I don't see the usecase of having different sanitization options for `Segment` and its subelements. Probably one parameter is sufficient. The most consistent name would be `sanitize_names`.

cheetah/accelerator/segment.py
```diff
@@ -319,18 +332,21 @@ def from_ocelot(
...
-        return cls(converted, name=name, **kwargs)
+        return cls(converted, name=name, sanitize_name=sanitize_name, **kwargs)
 
     @classmethod
     def from_bmad(
         cls,
         bmad_lattice_file_path: str,
         environment_variables: dict | None = None,
-        warnings: bool = True,
+        sanitize_element_names: bool = False,
```
Hespe, line 349: Should be consistent with the other import functions

cheetah/converters/bmad.py
```diff
@@ -10,18 +11,22 @@
     read_clean_lines,
     validate_understood_properties,
 )
+from cheetah.utils import PhysicsWarning
 
 
 def convert_element(
     name: str,
     context: dict,
+    sanitze_name: bool = False,
```
Hespe, line 20: typo

cheetah/converters/bmad.py
```diff
@@ -38,10 +43,11 @@ def convert_element(
     if isinstance(bmad_parsed, list):
         return cheetah.Segment(
             elements=[
-                convert_element(element_name, context, device, dtype)
+                convert_element(element_name, context, sanitze_name, device, dtype)
                 for element_name in bmad_parsed
             ],
             name=name,
+            sanitize_name=sanitze_name,
```
Hespe, line 50: Is the top-level element after import always a `Segment`? In that case, we could skip sanitizing all lower level names.

cheetah/converters/bmad.py
```diff
@@ -298,7 +332,7 @@ def convert_element(
 def convert_lattice_to_cheetah(
     bmad_lattice_file_path: Path,
     environment_variables: dict | None = None,
-    warnings: bool = True,
+    sanitize_element_names: bool = False,
```
Hespe, line 335: Is there a reason for using a different variable name here?

cheetah/converters/elegant.py
```diff
@@ -37,17 +43,20 @@ def convert_element(
     if isinstance(parsed, list):
         return cheetah.Segment(
             elements=[
-                convert_element(element_name, context, device, dtype)
+                convert_element(element_name, context, sanitize_name, device, dtype)
                 for element_name in parsed
             ],
             name=name,
+            sanitize_name=sanitize_name,
```
Hespe, line 50: Same as for bmad, do we need to sanitize all subelements?

PR #452, opened 2025-06-02
Optimize transfer map calculation of rotated elements

Hespe:
## Description

Rotation matrices are always orthogonal. We therefore do not have to invert or recompute them to obtain the rotation in the opposite direcition but rather just transpose. This PR changes the transfer map calculation of rotated elements to take advantage of this, which should result in a small speedup.

## Motivation and Context

## Types of changes

- [x] Speed optimization

## Checklist

### DIFF
--- a/cheetah/accelerator/dipole.py
+++ b/cheetah/accelerator/dipole.py
@@ -414,7 +414,12 @@ class Dipole(Element):
         R = R_exit @ R @ R_enter
 
         # Apply rotation for tilted magnets
-        R = rotation_matrix(-self.tilt) @ R @ rotation_matrix(self.tilt)
+        if torch.any(self.tilt != 0):
+            rotation = rotation_matrix(self.tilt)
+
+            # Rotation matrices are orthogonal. We can therefore transpose the matrix to
+            # rotate in the other direction.
+            R = rotation.transpose(-1, -2) @ R @ rotation
 
         return R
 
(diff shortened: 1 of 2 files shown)


cheetah/accelerator/dipole.py
```diff
@@ -414,7 +414,12 @@ def transfer_map(self, energy: torch.Tensor, species: Species) -> torch.Tensor:
         R = R_exit @ R @ R_enter
 
         # Apply rotation for tilted magnets
-        R = rotation_matrix(-self.tilt) @ R @ rotation_matrix(self.tilt)
+        if torch.any(self.tilt != 0):
+            rotation = rotation_matrix(self.tilt)
+
+            # Rotation matrices are orthogonal. We can therefore transpose the matrix to
+            # rotate in the other direction.
+            R = rotation.transpose(-1, -2) @ R @ rotation
```
jank324, line 422:
```suggestion
        if torch.any(self.tilt != 0):
            rotation = rotation_matrix(self.tilt)
            R = rotation.transpose(-1, -2) @ R @ rotation
```

jank324 (changes requested):
Can you add the changelog entry and fill in the tickboxes?

jank324:
I also removed the last `,` in the `einsum` call because I think black might now format this onto a single line.

Hespe:
> I also removed the last `,` in the `einsum` call because I think black might now format this onto a single line.

I've checked locally, black is just adding back the `,` in this case.

jank324 (approved):

PR #463, opened 2025-06-11
Define minimum versions for all dependencies

jank324:
## Description

Adds minimum versions to all dependencies, and a test case that runs the tests with those minimum requirements. This should catch when we use features that aren't available in old versions of the dependencies, and ensure that users don't accidentally install Cheetah with other packages that are not compatible.

## Motivation and Context

We ran into this issue with an old version of PyTorch (`2.2.2`) that caused Cheetah to throw somewhat cryptic errors.

## Types of changes

## Checklist


Hespe:
The extra test run would have caught the incompatibility with pytorch 2.2.

.github/workflows/pytest.yaml
```diff
@@ -12,11 +12,17 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         python-version: ["3.10", "3.11", "3.12", "3.13"]
+        constraints: [""]
         include:
+          - os: ubuntu-latest
+            python-version: "3.10"
+            constraints: "-c test_constraints.txt"
```
jank324, line 19: Actually ... this could be `options`?

PR #471, opened 2025-06-12
Improve the cov calculation in ParticleBeam to alllow batched calculation

### LINKED ISSUE #470: `ParticleBeam.as_parameter_beam()` breaks for batched ParticleBeam
When ParticleBeam is batched, calling `as_parameter_beam` will throw an error

```python

import cheetah
import torch

# works
particle_beam = cheetah.ParticleBeam.from_parameters(num_particles=1000, mu_x=torch.tensor(1e-3),energy=torch.tensor(1e9), dtype=torch.float64)
param_beam = particle_beam.as_parameter_beam()

# this breaks
batched_particle_beam = cheetah.ParticleBeam.from_parameters(num_particles=1000, mu_x=torch.tensor((1e-3, 2e-3)),energy=torch.tensor(1e9), dtype=torch.float64)
param_beam = batched_particle_beam.as_parameter_beam()
# RuntimeError: cov(): expected input to have two or fewer dimensions but got an input with 3 dimensions
```

cr-xu:
## Description

Rework the covariance computation in `ParticleBeam.as_parameter_beam` to fix an issue that caused the covariance to be computed incorrectly for vectorised beams.

## Motivation and Context

Closes #470.

## Types of changes

## Checklist

### DIFF
--- a/cheetah/particles/particle_beam.py
+++ b/cheetah/particles/particle_beam.py
@@ -1041,7 +1041,7 @@ class ParticleBeam(Beam):
             mu=(self.particles * self.survival_probabilities.unsqueeze(-1)).sum(dim=-2)
             / self.survival_probabilities.sum(),
             cov=unbiased_weighted_covariance_matrix(
-                self.particles, weights=self.survival_probabilities, dim=-2
+                self.particles, weights=self.survival_probabilities
             ),
             energy=self.energy,
             total_charge=self.total_charge,
--- a/cheetah/particles/particle_beam.py
+++ b/cheetah/particles/particle_beam.py
@@ -555,7 +555,7 @@ class ParticleBeam(Beam):
             dtype=dtype,
         )
 
-        # Extract the batch dimension of the beam
+        # Extract the vector dimension of the beam
         vector_shape = beam.sigma_x.shape
 
         # Generate random particles in unit sphere in polar coodinates
@@ -858,7 +858,7 @@ class ParticleBeam(Beam):
         NOTE: openPMD uses boolean particle status flags, i.e. alive or dead. Cheetah's
             survival probabilities are converted to status flags by thresholding at 0.5.
 
-        NOTE: At the moment this method only supports non-batched particles
+        NOTE: At the moment this method only supports non-vectorised particle
             distributions.
 
         :return: openPMD `ParticleGroup` object with the `ParticleBeam`'s particles.
@@ -871,9 +871,11 @@ class ParticleBeam(Beam):
                 installed."""
             )
 
-        # For now only support non-batched particles
+        # For now only support non-vectorised particle distributions
         if len(self.particles.shape) != 2:
-            raise ValueError("Only non-batched particles are supported.")
+            raise ValueError(
+                "Only non-vectorised particle distributions are supported."
+            )
 
         px = self.px * self.p0c
         py = self.py * self.p0c
--- a/cheetah/particles/particle_beam.py
+++ b/cheetah/particles/particle_beam.py
@@ -1039,7 +1039,9 @@ class ParticleBeam(Beam):
 
         return ParameterBeam(
             mu=self.particles.mean(dim=-2),
-            cov=self.cov_matrix,
+            cov=unbiased_weighted_covariance_matrix(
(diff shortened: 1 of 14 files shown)


jank324 (approved):

cheetah/particles/particle_beam.py
```diff
@@ -1036,7 +1039,9 @@ def as_parameter_beam(self) -> "ParameterBeam":  # noqa: F821
 
         return ParameterBeam(
             mu=self.particles.mean(dim=-2),
-            cov=torch.cov(self.particles.transpose(-2, -1)),
+            cov=unbiased_weighted_covariance_matrix(
+                self.particles, weights=self.survival_probabilities, dim=-2
```
Hespe, line 1043: If we weigh the covariance with the survival probability (which I agree with and think is the right thing to do), we should do the same for the mean.

cheetah/utils/statistics.py
```diff
@@ -60,3 +60,25 @@ def unbiased_weighted_std(
...
+
+def unbiased_weighted_covariance_matrix(
+    inputs: torch.Tensor, weights: torch.Tensor, dim: int = -2
+) -> torch.Tensor:
+    """
+    Compute the unbiased weighted covariance matrix of a tensor.
+
+    :param inputs: Input tensor of shape (..., sample_size, n_features).
+    :param weights: Weights tensor of shape (..., sample_size).
+    :param dim: Dimension along which to compute the covariance.
```
Hespe, line 73: I am not sure that the `dim` argument actually works for anything but `-2`. The final `matmul` always computes the product with the last two dimensions.

cheetah/particles/particle_beam.py
jank324: I added that.

cheetah/utils/statistics.py
jank324: So you think I did break it? I guess we should include that in the test. 

The test also fails sporadically. I reduced the number of particles from 10 million to 1 million because the test took so long to run. It seems that that does play nicely with the stochasticity and tolerances though.

cheetah/utils/statistics.py
```diff
@@ -60,3 +60,25 @@ def unbiased_weighted_std(
...
+    :return: Unbiased weighted covariance matrix.
+    """
+    normalized_weights = weights / torch.sum(weights)
+    weighted_means = torch.sum(inputs * normalized_weights.unsqueeze(-1), dim)
+    centered_inputs = inputs - weighted_means.unsqueeze(dim)
+    correction_factor = 1 - (torch.sum(normalized_weights**2) / torch.sum(weights))
```
Hespe, line 79: And I think the correction factor does not match the scalar version of
https://github.com/desy-ml/cheetah/blob/821d2e2b14dc668416b9a1d66b0999f3384ac4ce/cheetah/utils/statistics.py#L18-L20

There we do not normalize the weights first, but the mixing of normalized and given weights seems off.

cheetah/utils/statistics.py
Hespe: I think the `dim` argument was broken before but `torch.sum(weights)` now normalizes over all dimensions while before it did so batchwise. And batchwise is likely what we need since the batches might e.g. be different apertures with very different survival probabilities.

jank324: So the weights sum is now similar to how it was before, and I vectorised the `survival_probabilities in the test, which should cause `weights` to be vectorised. Now everything is broken. 😂

I also checked more beam parameters in the test now, and added an absolute tolerance to all the mus that I think should be there.

cr-xu: I agree, I put the dim argument there just trying to be consistent with other utility methods of calculating weighted mean and cov, but right now it's just working for dim=-2. I can also remove that.

jank324 (changes requested):
Actually, we realised this doesn't work yet.

Hespe:
The covariance matrix calculation now passes the isolated tests, also with vectorized weights. The beam roundtrip test still fails but this maybe a question of proper tolerances.

jank324 (approved):

PR #469, opened 2025-06-12
Increase the precision of base_rmatrix calculation to double precision

cr-xu:
## Description

Update the handling of small k1 values so that the `base_rmatrix` produces smaller error.

## Motivation and Context

- [x] I have raised an issue to propose this change (required for new features and bug fixes)

Closes #468.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation (update in the documentation)

## Checklist

- [x] I have updated the changelog accordingly (**required**).
- [ ] My change requires a change to the documentation.
- [x] I have updated the tests accordingly (_required for a bug fix or a new feature_).
- [ ] I have updated the documentation accordingly.
  
  
- [x] I have reformatted the code and checked that formatting passes (**required**).
- [x] I have have fixed all issues found by `flake8` (**required**).
- [x] I have ensured that all `pytest` tests pass (**required**).
- [ ] I have run `pytest` on a machine with a CUDA GPU and made sure all tests pass (**required**).
- [x] I have checked that the documentation builds (**required**).

Note: We are using a maximum length of 88 characters per line.

PR #474, opened 2025-06-16
Improve numerical precision of inactive elements

### LINKED ISSUE #472: Use `torch.sinc` in `base_rmatrix`
As a better fix for numerical issues, which we tried to address in #469, use `torch.sinc` in `base_rmatrix` to replace expressions like

```python
sy = (torch.sin(ky * length) / ky).real
```

and also let inactive elements (e.g. quadruples) use drift matrices when they are turned off.

jank324:
## Description

Inactive elements, like quadrupoles with `k1 == 0.0` now catch these cases in a way that avoids numerical errors. For example, instead of avoid a division by 0.0, by replacing `k1` with a very small positive value, these cases are caught more properly where the division by `0.0` is performed.

Additionally adds a minor speed optimisation in `base_rmatrix` and `base_ttensor` that reuses a tensor with constant value to save some memory allocations.

## Motivation and Context

Closes #472. Improves upon #469.

## Types of changes

## Checklist


jank324:
I'm guessing the same probably needs to be done for `base_ttensor`.

PR #476, opened 2025-06-18
Add second order tracking to more elements

cr-xu:
- Added a general `track_second_oder` method in 
- Add `second_order` as additional `tracking_method` for drift, dipole, and quadrupole. If selected, element will use second order tracking.

TODO:

- [ ] Proper handling of tilt (rotation) and misalignment for second order tracking.
- [ ] Add a convenient method to set the segment element tracking methods

## Motivation and Context

## Types of changes

## Checklist

### DIFF
--- a/cheetah/accelerator/dipole.py
+++ b/cheetah/accelerator/dipole.py
@@ -6,7 +6,7 @@ from matplotlib.patches import Rectangle
 
 from cheetah.accelerator.element import Element
 from cheetah.particles import Beam, ParticleBeam, Species
-from cheetah.track_methods import base_rmatrix, rotation_matrix
+from cheetah.track_methods import base_rmatrix, base_ttensor, rotation_matrix
 from cheetah.utils import UniqueNameGenerator, bmadx, verify_device_and_dtype
 
 generate_unique_name = UniqueNameGenerator(prefix="unnamed_element")
@@ -57,7 +57,7 @@ class Dipole(Element):
         fringe_integral_exit: torch.Tensor | None = None,
         fringe_at: Literal["neither", "entrance", "exit", "both"] = "both",
         fringe_type: Literal["linear_edge"] = "linear_edge",
-        tracking_method: Literal["cheetah", "bmadx"] = "cheetah",
+        tracking_method: Literal["cheetah", "bmadx", "second_order"] = "cheetah",
         name: str | None = None,
         sanitize_name: bool = False,
         device: torch.device | None = None,
@@ -176,6 +176,8 @@ class Dipole(Element):
         """
         if self.tracking_method == "cheetah":
             return super().track(incoming)
+        elif self.tracking_method == "second_order":
+            return super().track_second_order(incoming)
         elif self.tracking_method == "bmadx":
             assert isinstance(
                 incoming, ParticleBeam
@@ -423,6 +425,19 @@ class Dipole(Element):
             R = rotation.transpose(-1, -2) @ R @ rotation
 
         return R
+
+    def second_order_map(self, energy: torch.Tensor, species: Species) -> torch.Tensor:
+        T = base_ttensor(
+            length=self.length,
+            k1=torch.zeros_like(self.length),
+            k2=torch.zeros_like(self.length),
+            hx=self.hx,
+            species=species,
+            tilt=self.tilt,
+            energy=energy,
+        )
+
+        return T
 
     def _transfer_map_enter(self) -> torch.Tensor:
         """Linear transfer map for the entrance face of the dipole magnet."""
--- a/cheetah/accelerator/drift.py
+++ b/cheetah/accelerator/drift.py
@@ -5,6 +5,7 @@ import torch
 
 from cheetah.accelerator.element import Element
 from cheetah.particles import Beam, ParticleBeam, Species
+from cheetah.track_methods import base_ttensor
 from cheetah.utils import (
     UniqueNameGenerator,
     bmadx,
@@ -33,7 +34,7 @@ class Drift(Element):
     def __init__(
         self,
         length: torch.Tensor,
-        tracking_method: Literal["cheetah", "bmadx"] = "cheetah",
+        tracking_method: Literal["cheetah", "bmadx", "second_order"] = "cheetah",
         name: str | None = None,
         sanitize_name: bool = False,
         device: torch.device | None = None,
@@ -62,6 +63,17 @@ class Drift(Element):
 
         return tm
 
+    def second_order_map(self, energy: torch.Tensor, species: Species) -> torch.Tensor:
+        T = base_ttensor(
+            self.length,
+            k1=torch.zeros_like(self.length),
+            k2=torch.zeros_like(self.length),
+            hx=torch.zeros_like(self.length),
+            energy=energy,
+            species=species,
+        )
+        return T
+
     def track(self, incoming: Beam) -> Beam:
         """
         Track particles through the dipole element.
@@ -71,6 +83,8 @@ class Drift(Element):
         """
         if self.tracking_method == "cheetah":
             return super().track(incoming)
+        elif self.tracking_method == "second_order":
+            return super().track_second_order(incoming)
         elif self.tracking_method == "bmadx":
             assert isinstance(
                 incoming, ParticleBeam
--- a/cheetah/accelerator/element.py
+++ b/cheetah/accelerator/element.py
@@ -7,7 +7,11 @@ import torch
 from torch import nn
 
 from cheetah.particles import Beam, ParameterBeam, ParticleBeam, Species
-from cheetah.utils import DirtyNameWarning, UniqueNameGenerator
+from cheetah.utils import (
+    DirtyNameWarning,
+    NotSupportedTrackingMethodWarning,
+    UniqueNameGenerator,
+)
 
 generate_unique_name = UniqueNameGenerator(prefix="unnamed_element")
 
@@ -76,6 +80,19 @@ class Element(ABC, nn.Module):
         """
         raise NotImplementedError
 
+    def second_order_map(self, energy: torch.Tensor, species: Species) -> torch.Tensor:
+        r"""
+        Generates the element's second-order transfer map that describes how the beam
+        and its particles are transformed when traveling through the element.
+
+        :math:`pout_{i} = \sum_{j,k} T_{ijk} pin_{j} pin_{k}`
+
+        :param energy: Reference energy of the beam. Read from the fed-in Cheetah beam.
+        :param species: Species of the particles in the beam
+        :return: A 7x7x7 Tensor T_ijk for further calculations.
+        """
+        raise NotImplementedError
+
     def track(self, incoming: Beam) -> Beam:
         """
         Track particles through the element. The input can be a `ParameterBeam` or a
@@ -111,6 +128,45 @@ class Element(ABC, nn.Module):
             )
         else:
             raise TypeError(f"Parameter incoming is of invalid type {type(incoming)}")
+
+    def track_second_order(self, incoming: Beam):
+        """
+        Track particles through the element with second-order transfer maps.
+        The input can be a `ParameterBeam` or a `ParticleBeam`.
+
+        :param incoming: Beam of particles entering the element.
+        :return: Beam of particles exiting the element.
+        """
+        if isinstance(incoming, ParameterBeam):
+            warnings.warn(
+                "Second order tracking is not supported for `ParameterBeam` "
+                "for now. Falling back to first-order tracking instead.",
+                category=NotSupportedTrackingMethodWarning,
+                stacklevel=2,
+            )
+            return self.track(incoming)
+
+        first_order_tm = self.transfer_map(incoming.energy, incoming.species)
+        second_order_tm = self.second_order_map(incoming.energy, incoming.species)
+
+        # Apply the transfer map to the incoming particles
+        first_order_particles = incoming.particles @ first_order_tm.transpose(-2, -1)
+        second_order_particles = torch.einsum(
+            "...ijk,...j,...k->...i",
+            second_order_tm.unsqueeze(-4),  # Add broadcast dimension for particles
+            incoming.particles,
+            incoming.particles,
+        )
+        outgoing_particles = second_order_particles + first_order_particles
+
+        return ParticleBeam(
+            particles=outgoing_particles,
+            energy=incoming.energy,
+            particle_charges=incoming.particle_charges,
+            survival_probabilities=incoming.survival_probabilities,
+            s=incoming.s + self.length,
+            species=incoming.species,
+        )
 
     def forward(self, incoming: Beam) -> Beam:
         """Forward function required by `torch.nn.Module`. Simply calls `track`."""
--- a/cheetah/accelerator/sextupole.py
+++ b/cheetah/accelerator/sextupole.py
@@ -3,7 +3,7 @@ import torch
 from matplotlib.patches import Rectangle
 
 from cheetah.accelerator.element import Element
-from cheetah.particles import Beam, ParameterBeam, ParticleBeam, Species
+from cheetah.particles import Beam, Species
 from cheetah.track_methods import base_rmatrix, base_ttensor, misalignment_matrix
 from cheetah.utils import squash_index_for_unavailable_dims, verify_device_and_dtype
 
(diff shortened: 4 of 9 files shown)


cr-xu:
I just realized that there's already a method called `set_attrs_on_every_element_of_type()`, which can be used to set `tracking_method` to `second_order`, but one needs to specify all the element types, but one can easily miss something...

Hespe:
> Right now drift, dipole, quadrupole have `return self.tracking_method == "cheetah"` to check whether they a skippable, which is correct.

Yes, you are right, I've missed that. So no need to change anything if we don't want to.

cheetah/accelerator/drift.py
```diff
@@ -62,15 +68,36 @@ def transfer_map(self, energy: torch.Tensor, species: Species) -> torch.Tensor:
 
         return tm
 
+    def second_order_map(self, energy: torch.Tensor, species: Species) -> torch.Tensor:
+        T = base_ttensor(
```
Hespe, line 72: For `transfer_map`, `Drift` implements the raw map itself rather than calling `base_rmatrix` with 0 coefficients. Should we be consistent there (in either way)?

cheetah/accelerator/element.py
```diff
@@ -112,6 +129,45 @@ def track(self, incoming: Beam) -> Beam:
...
+
+        # Apply the transfer map to the incoming particles
+        first_order_particles = incoming.particles @ first_order_tm.transpose(-2, -1)
+        second_order_particles = torch.einsum(
+            "...ijk,...j,...k->...i",
+            second_order_tm.unsqueeze(-4),  # Add broadcast dimension for particles
+            incoming.particles,
+            incoming.particles,
+        )
+        outgoing_particles = second_order_particles + first_order_particles
```
Hespe, line 160: We could of course apply the same trick we did for the constant part of the map to the first-order part. That is, merge it into the `T` tensor by using the constant 1 of the 7-dimensional phase space. That would allow us to at least merge uninterrupted linear maps into a second-order one that follows immediatly.

cheetah/accelerator/segment.py
```diff
@@ -566,6 +566,17 @@ def set_attrs_on_every_element_of_type(
...
+    def set_tracking_method(
+        self,
+        tracking_method: Literal[
+            "linear", "cheetah", "bmadx", "second_order"
+        ] = "linear",
+    ) -> None:
+        for element in self.elements:
+            if hasattr(element, "tracking_method"):
+                if tracking_method in element.supported_tracking_methods:
+                    element.tracking_method = tracking_method
```
Hespe, line 578: I am not sure if simply ignoring elements that do not support that particular tracking method is the right call. Maybe we need some kind of default tracking method we set elements to instead?

cheetah/accelerator/element.py
```diff
@@ -112,6 +129,45 @@ def track(self, incoming: Beam) -> Beam:
...
+        :return: Beam of particles exiting the element.
+        """
+        if isinstance(incoming, ParameterBeam):
+            warnings.warn(
+                "Second order tracking is not supported for `ParameterBeam` "
+                "for now. Falling back to first-order tracking instead.",
+                category=NotSupportedTrackingMethodWarning,
+                stacklevel=2,
+            )
+            return Element.track(self, incoming)
```
Hespe, line 147: For `bmadx` tracking, we do not fall back with a warning but throw an `AssertionError`. I think we should be consistent in how we handle incompatible `tracking_method` and `Beam`.

cheetah/utils/__init__.py
```diff
@@ -20,6 +20,7 @@
     DefaultParameterWarning,
     DirtyNameWarning,
     NoBeamPropertiesInLatticeWarning,
+    NotSupportedTrackingMethodWarning,
```
Hespe, line 23: #497 made warnings available in the top level namespace. We should add this one there as well.

tests/test_segment.py
```diff
@@ -171,3 +171,28 @@ def test_longitudinal_beam_metric(attr_names):
...
+    """
+
+    segment = cheetah.Segment(
+        elements=[
+            cheetah.Drift(length=torch.tensor(0.5), name="d1"),
+            cheetah.Quadrupole(length=torch.tensor(0.3), name="q1"),
+            cheetah.Drift(length=torch.tensor(0.2), name="d2"),
+            cheetah.Dipole(length=torch.tensor(0.5), name="b1"),
+        ]
+    )
```
Hespe, line 192: I think the `Segment` should contain an `Element` which does not support all of the tracking methods.

tests/test_sextupole.py
```diff
@@ -136,3 +136,33 @@ def test_sextupole_as_drift():
...
+    )
+    outbeam_sextupole_with_misalignment = sextupole_with_misalignment(incoming_beam)
+    outbeam_sextupole_without_misalignment = sextupole_without_misalignment(
+        incoming_beam
+    )
+
+    assert not torch.allclose(
+        outbeam_sextupole_with_misalignment.mu_x,
+        outbeam_sextupole_without_misalignment.mu_x,
+    )
```
Hespe, line 168: Instead of just testing that a misalign and aligned `Sextupole` behave differently, we should be able to compare the misalign magnet to an aligned one that tracks a shifted `Beam`, right?

cheetah/accelerator/element.py
cr-xu: This is now replaced with an assertion error to be consistent with bmadx

tests/test_sextupole.py
cr-xu: Changed as suggested.

cheetah/accelerator/element.py
```diff
@@ -112,6 +129,45 @@ def track(self, incoming: Beam) -> Beam:
         else:
             raise TypeError(f"Parameter incoming is of invalid type {type(incoming)}")
 
+    def track_second_order(self, incoming: Beam):
```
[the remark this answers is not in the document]
cr-xu: I think this can be implemented in a later PR. It would involve modifying the cavity, correctors, and other elements

cheetah/accelerator/drift.py
cr-xu: I think this was done earlier when transfer_map would introduce a small error when calling with k1=0. It's probably not relevant anymore as we replaced that with `sinc`. This can be addressed in a separate issue though. 

For the second order map I don't see a problem with calling the base_ttensor function, having a custom simpliefied t_tensor calculation could be minimally faster, but not sure if it's worth it.

cr-xu:
Ping @Hespe  @jank324, is something else to be fixed for this PR or we can proceed?

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 - Add the ability to the Bmad and Elegant converters to parse expressions that access properties from other elements (see #501, #498) (@amylizzle, @jank324)
 - Update the Elegant converter to allow element names with colon as well as the definition of reversed beamlines with a minus sign. (see #504) (@cr-xu, @jank324)
 - `Segment`s can now conveniently be reversed with the `Segment.reversed` method (see #504) (@jank324)
+- Allow second-order tracking for dipole, drift, and quadrupole elements, and add a convenient method to set tracking method for all the segment. Rename the tracking method `cheetah` to `linear`. (see #476) (@cr-xu)
```
jank324, line 14:
```suggestion
- Allow second-order tracking for `Drift`, `Dipole` and `Quadrupole` elements, and add a convenient method to set tracking methods for an entire segment. Rename the tracking method `"cheetah"` to `"linear"`. `"cheetah"` will remain supported with a `DeprecationWarning`. (see #476) (@cr-xu)
```

jank324: Is this entry still correct?

cheetah/accelerator/dipole.py
```diff
@@ -43,6 +44,8 @@ class Dipole(Element):
         syntax to access the element in a segment.
     """
 
+    supported_tracking_methods = ["linear", "cheetah", "bmadx", "second_order"]
```
jank324, line 47: Would there be a better name for `"bmadx"` as well? It would only be consistent to deprecate both `"cheetah"` and `"bmadx"` as tracking method names.

cheetah/accelerator/dipole.py
```diff
@@ -161,7 +166,7 @@ def dipole_e2(self, value: torch.Tensor):
 
     @property
     def is_skippable(self) -> bool:
-        return self.tracking_method == "cheetah"
+        return self.tracking_method == "cheetah" or self.tracking_method == "linear"
```
jank324, line 169:
```suggestion
        return self.tracking_method in ["linear", "cheetah"]
```

cheetah/accelerator/dipole.py
```diff
@@ -174,8 +179,18 @@ def track(self, incoming: Beam) -> Beam:
         :param incoming: Beam entering the element.
         :return: Beam exiting the element.
         """
-        if self.tracking_method == "cheetah":
+        if self.tracking_method == "linear":
+            return super().track(incoming)
```
jank324, line 183: Should we also have a `track_first_order` method to make this a bit clearer?

In the future we might have to think about replacing branching with function variables for the sake of speed ... this would lay the groundwork for that ... I guess (?)

cheetah/accelerator/dipole.py
```diff
@@ -424,6 +439,19 @@ def transfer_map(self, energy: torch.Tensor, species: Species) -> torch.Tensor:
 
         return R
 
+    def second_order_map(self, energy: torch.Tensor, species: Species) -> torch.Tensor:
```
jank324, line 442:
```suggestion
    def second_order_transfer_map(self, energy: torch.Tensor, species: Species) -> torch.Tensor:
```

cheetah/accelerator/dipole.py
```diff
@@ -424,6 +439,19 @@ def transfer_map(self, energy: torch.Tensor, species: Species) -> torch.Tensor:
...
+        T = base_ttensor(
+            length=self.length,
+            k1=torch.zeros_like(self.length),
+            k2=torch.zeros_like(self.length),
```
jank324, line 446:
```suggestion
            k1=torch.tensor(0.0, dtype=self.length.dtype, device=self.length.device),
            k2=torch.tensor(0.0, dtype=self.length.dtype, device=self.length.device),
```
We only need the device and dtype from `length`. If we do `zeros_like`, we would also take over the size, which might waste a lot of memory if `length` is vectorised.

cheetah/accelerator/transverse_deflecting_cavity.py
```diff
@@ -34,6 +34,8 @@ class TransverseDeflectingCavity(Element):
         syntax to access the element in a segment.
     """
 
+    supported_tracking_methods = ["bmadx"]
```
jank324, line 37: I like this! Should avoid confusion in the future. 😊

cheetah/track_methods.py
```diff
@@ -6,31 +6,30 @@
 from cheetah.utils import compute_relativistic_factors
 
 
-def base_rmatrix(
+def _base_rmatrix(
+    R: torch.Tensor,
```
jank324, line 10: I don't like this. Why are we doing that?

tests/test_compare_ocelot.py
```diff
@@ -10,30 +10,42 @@
...
     # Ocelot
     incoming_p_array = ocelot.astraBeam2particleArray(
         "tests/resources/ACHIP_EA1_2021.1351.001"
     )
     ocelot_bend = ocelot.Bend(l=0.1, angle=0.1)
-    lattice = ocelot.MagneticLattice([ocelot_bend])
+    if tracking_method == "second_order":
+        ocelot_method = {"global": ocelot.SecondTM}
+    else:
+        ocelot_method = {"global": ocelot.TransferMap}
```
jank324, line 39:
```suggestion
    ocelot_method = {"global" ocelot.SecondTM if tracking_method == "second_order" else ocelot.TransferMap}
```

cheetah/accelerator/dipole.py
cr-xu: I agree we should rename bmadx tracking emthod, right now I just don't find a proper name. Maybe `kick_drift_kick` or `hamiltonian`? It's not very self-explaining as linear and second_order though. In bmad it's just called `bmad_standard` which is also not helping.

cheetah/accelerator/dipole.py
cr-xu: Done!

jank324:
Tests are failing on all PRs as a result of an Ocelot update. See #512 for issue and #513 for PR to fix it. Currently blocking this PR and #511.

cheetah/accelerator/dipole.py
jank324: I like both. I think `kick_drift_kick` sounds more explanatory to me as a non-accelerator physicist. `hamiltonian` sounds cool and is nice and short. Maybe choose whichever seems more intuitive to accelerator physicists. I think we won't get around documenting their meaning. There we could reference the Bmad manual and `bmad_standard` method.

cheetah/accelerator/custom_transfer_map.py
```diff
@@ -21,10 +23,13 @@ class CustomTransferMap(Element):
         syntax to access the element in a segment.
     """
 
+    supported_tracking_methods = ["linear"]
+
     def __init__(
         self,
         predefined_transfer_map: torch.Tensor,
         length: torch.Tensor | None = None,
+        tracking_method: Literal["linear"] = "linear",
```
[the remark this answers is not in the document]
Hespe: Not without further adjustments. Adding `second_order` would not be difficult though. But that is probably best left for another PR.

tests/test_segment.py
```diff
@@ -171,3 +171,34 @@ def test_longitudinal_beam_metric(attr_names):
             assert attr_result.shape == (
                 (4, incoming_beam.num_particles) if attr_name == "x" else (4,)
             )
+
+
+@pytest.mark.parametrize(
+    "tracking_method",
+    ["linear", "second_order", "bmadx"],
+)
```
jank324, line 179:
```suggestion
@pytest.mark.parametrize(
    "tracking_method", ["linear", "second_order", "bmadx"]
)
```

cheetah/accelerator/horizontal_corrector.py
```diff
@@ -27,10 +29,13 @@ class HorizontalCorrector(Element):
         syntax to access the element in a segment.
     """
 
+    supported_tracking_methods = ["linear"]
+
     def __init__(
         self,
         length: torch.Tensor,
         angle: torch.Tensor | None = None,
+        tracking_method: Literal["linear"] = "linear",
```
jank324, line 38: Why add this argument, if this element only supports one tracking method?

cheetah/accelerator/marker.py
```diff
@@ -29,7 +29,9 @@ def __init__(
             name=name, sanitize_name=sanitize_name, device=device, dtype=dtype
         )
 
-    def transfer_map(self, energy: torch.Tensor, species: Species) -> torch.Tensor:
+    def first_order_transfer_map(
```
jank324, line 32: So with the tracking method change, `Marker` say it supports no tracking method. `Marker.track` returns a clone of the original beam, which is correct for any tracking method. 

**Note:** Elements can implement their own tracking methods, so just adding a list of supported tracking methods and listing all of them is not really correct either I think.

How should we handle this?

cheetah/accelerator/custom_transfer_map.py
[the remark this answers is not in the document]
jank324: That raises the question of why we have the argument now.

tests/test_sextupole.py
```diff
@@ -136,3 +140,39 @@ def test_sextupole_as_drift():
...
+
+    centered_incoming_beam = cheetah.ParticleBeam.from_parameters(
+        mu_x=torch.tensor(0.0),
+        sigma_px=torch.tensor(2e-7),
+        sigma_py=torch.tensor(2e-7),
+        sigma_p=torch.tensor(1e-2),
+    )
+    misaligned_incoming_beam = centered_incoming_beam.clone()
+    misaligned_incoming_beam.particles[:, 0] -= 1e-3
```
jank324, line 164: Why does this not do the same as
```python
misaligned_incoming_beam = centered_incoming_beam.transformed_to(mu_x=torch.tensor(0.0))
```

cheetah/accelerator/sextupole.py
```diff
@@ -16,18 +18,24 @@ class Sextupole(Element):
     :param k2: Sextupole strength in 1/m^3.
     :param misalignment: Transverse misalignment in x and y directions in meters.
     :param tilt: Tilt angle of the quadrupole in x-y plane in radians.
+    :param tracking_method: Method to use for tracking through the element.
+        Note: By default, the sextupole is created with linear tracking method so it
+        will not have second order effects.
```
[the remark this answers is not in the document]
cr-xu: I have no opinion here... I can set it back to default to 2nd_order

cheetah/accelerator/element.py
```diff
@@ -112,6 +177,44 @@ def track(self, incoming: Beam) -> Beam:
...
+        ), "Second-order tracking is currently only supported for `ParticleBeam`."
+
+        first_order_tm = self.first_order_transfer_map(
+            incoming.energy, incoming.species
+        )
+        second_order_tm = self.second_order_transfer_map(
+            incoming.energy, incoming.species
+        )
+        # Fill the first-order transfer map into the second-order transfer map.
+        second_order_tm[..., :, 6, :] = first_order_tm
```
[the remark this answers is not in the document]
cr-xu: I still prefer that there's a method to get the pure second order map. 
It could be confusing for benchmarking or debugging to have only the up_to_second_oder_map. It's also not unique, in the sense that we can insert the first order to T[:,6,:] or T[:,:,6].

cheetah/accelerator/sextupole.py
```diff
@@ -55,71 +63,63 @@ def __init__(
...
             tilt=self.tilt,
             energy=energy,
         )
 
-        if torch.all(self.misalignment == 0):
-            return R
-        else:
+        if not torch.all(self.misalignment == 0):
             R_entry, R_exit = misalignment_matrix(self.misalignment)
             R = R_exit @ R @ R_entry
```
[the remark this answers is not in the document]
jank324: Let's replace `base_rmatrix` here with a new `base_drift_matrix` or something like it.

jank324:
> If I am not mistaken, this PR should also close #515, right?

I think so.

PR #513, opened 2025-07-18
Fix Ocelot-related tests failing

### LINKED ISSUE #512: Tests involving Ocelot are failing
There seems to be a new version of Ocelot that causes tests comparing to it to fail with
```
AttributeError: property 'gamma_x' of 'Twiss' object has no setter
```

jank324:
## Description

Changes the Ocelot dependency from the `master` branch of the Ocelot GitHub repository to its new PyPI release. Fixes tests that use old interface of Ocelot, and increases tolerances on tests where needed.

## Motivation and Context

Closes #512.

## Types of changes

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,6 @@
 - Update the Elegant converter to allow element names with colon as well as the definition of reversed beamlines with a minus sign. (see #504) (@cr-xu, @jank324)
 - `Segment`s can now conveniently be reversed with the `Segment.reversed` method (see #504) (@jank324)
 - New feature for generating 3D models of lattices and viewing them (see #352, #502, #511) (@jank324, @chrisjcc, @SuchethShenoy)
-- Add a `twiss_parameter` method to compute the trace space Twiss parameters for `ParticleBeam` (see #513) (@cr-xu)
 
 ### 🐛 Bug fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 - Add a Binder and link to `cheetah-demos` (see #482) (@smartsammler, @jank324)
 - PyTorch is now configured to use only deterministic algorithms during tests, preventing intermittent test failures (see #480) (@Hespe)
 - Make README example more copy-paste friendly, and generally improve it and the simple intro notebook in the docs. (see #493, #496) (@jank324, @ax3l)
+- Fix comparison tests to work with new PyPI release of Ocelot (see #513) (@jank324)
 
 ### 🌟 First Time Contributors
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@
 - Add a Binder and link to `cheetah-demos` (see #482) (@smartsammler, @jank324)
 - PyTorch is now configured to use only deterministic algorithms during tests, preventing intermittent test failures (see #480) (@Hespe)
 - Make README example more copy-paste friendly, and generally improve it and the simple intro notebook in the docs. (see #493, #496) (@jank324, @ax3l)
-- Fix comparison tests to work with new PyPI release of Ocelot (see #513, #519) (@jank324)
+- Fix comparison tests to work with new PyPI release of Ocelot (see #513, #519) (@jank324, @cr-xu, @Hespe)
 
 ### 🌟 First Time Contributors
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Update the Elegant converter to allow element names with colon as well as the definition of reversed beamlines with a minus sign. (see #504) (@cr-xu, @jank324)
 - `Segment`s can now conveniently be reversed with the `Segment.reversed` method (see #504) (@jank324)
 - New feature for generating 3D models of lattices and viewing them (see #352, #502) (@jank324, @chrisjcc)
+- Add a `twiss_parameter` method to calculate the trace space twiss paramters for `ParticleBeam` (see #513) (@cr-xu)
 
 ### 🐛 Bug fixes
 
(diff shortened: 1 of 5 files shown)


jank324:
@cr-xu can you check if those tolerance increases seem plausible to you? These tests used to pass with much lower tolerances before, and some of the tolerances are no quite high at `trol=1e-2`. Some tests that this pertained to now already had `TODO` comments about the tolerances on them as well.

Could the removal of the `gamma_x` and `gamma_y` assignments be the cause of that, or an internal change in Ocelot?

cr-xu:
I'm looking into the issue. I thought it was related to dispersion but it's not... It's definitely out of the expected error range.

cr-xu:
The issue is that we are calculating different twiss values. Right now in cheetah we calculate twiss with canonical variables, and ocelot calculate twiss in trace space (which is also more common elsewhere). Both definitions are valid. I think I'll add the other variant into Cheetah as well. 
The question is just whether we want to leave the large errors and merge this right now to have the CI/CD passing for other PRs, since the proper introduction of twiss calculation fits better for a standalone PR.

cr-xu:
I added the calculation for `ParticleBeam` and removed the ocelot twiss test for ParameterBeam. 
The reason is that the transformation from canonical variables to trace space variables is non-linear, and we cannot properly transform the cov alone to get the correct twiss values, so the test does not make sense. 
It can be easily checked that if we have a ParameterBeam, and do `parameter_beam.as_particle_beam(num_particles).twiss_parameters` repeatedly, it will return twiss values with a few percent error each time.

cr-xu:
I think now the issue is fixed and can be merged. We can address more details on the twiss calculation in a separate PR.

cr-xu (approved):

cheetah/particles/beam.py
```diff
@@ -370,9 +370,63 @@ def relativistic_beta(self) -> torch.Tensor:
...
+            "Dx": Dx,
+            "Dxp": Dxp,
+            "beta_y": beta_y,
+            "alpha_y": alpha_y,
+            "gamma_y": gamma_y,
+            "emit_y": emit_y,
+            "Dy": Dy,
+            "Dyp": Dyp,
+        }
+        return twiss_parameters
```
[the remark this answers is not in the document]
jank324: See next comment on test, why that might also be unnecessary.

tests/test_compare_ocelot.py
```diff
@@ -311,72 +311,34 @@ def test_twiss_particle_beam():
...
+
     # Ocelot
     p_array = ocelot.astraBeam2particleArray(
         "tests/resources/ACHIP_EA1_2021.1351.001", print_params=False
     )
     ocelot_twiss = ocelot.cpbd.beam.get_envelope(p_array)
 
     # Compare
-    assert np.isclose(particle_beam.emittance_x.cpu().numpy(), ocelot_twiss.emit_x)
+    assert np.isclose(cheetah_twiss["emit_x"].cpu().numpy(), ocelot_twiss.emit_x)
```
jank324, line 323: So the thing is that the purpose of this test was to check whether property computations like `particle_beam.emittance_x` return correct results. By computing a different property and testing against that, we get the test to pass, but we also defeat its purpose.

jank324: I'm also still confused why we need the trace space coordinates now on the new release. With the old release, we passed with much better accuracy. What changed?

jank324:
I tracked down the change that made the Ocelot tests go from pretty much perfect accuracy to failing.

**Commit**: [92b5fd95ecbec63f094fbbf5f6361c43bff34600](https://github.com/ocelot-collab/ocelot/commit/92b5fd95ecbec63f094fbbf5f6361c43bff34600)

**Message:** "better approximation for moments calculation in get_envelope function"

Screenshot 2025-07-22 at 13 51 42

Specifically going from
```python
px = px * (1. - 0.5 * px * px - 0.5 * py * py)
```
to
```python
px = px * (1. - p - 0.5 * p * p + 0.5 * px * px + 0.5 * py * py )
```
and the same for `y` causes the tests to fail. Everything is fine with the first option, with the second we get a slight mismatch, though the general values are still similar.

jank324:
I didn't close it though.

jank324:
I didn't close it though.

jank324:
> > Some thoughts on this PR:
> > The `twiss_parameters` function should not exist alongside the normal Twiss parameter properties. If the Twiss parameters should be computed from trace space (for the `ParticleBeam` case), then those properties should do that (?)
> 
> Sure, it's just a temporary function. I think in the future this will be a flag/arugment that user can choose, whether to calculate the trace space (geometrical) twiss or the canonical one.

So a reasonable use case is that someone might want either of the two Twiss parameters? ... which would mean it would make sense to have a `ParticleBeam.geomeotric_beta_x` and a `ParticleBeam.canonical_beta_x`?

jank324:
> > > Some thoughts on this PR:
> > > The `twiss_parameters` function should not exist alongside the normal Twiss parameter properties. If the Twiss parameters should be computed from trace space (for the `ParticleBeam` case), then those properties should do that (?)
> > 
> > 
> > Sure, it's just a temporary function. I think in the future this will be a flag/arugment that user can choose, whether to calculate the trace space (geometrical) twiss or the canonical one.
> 
> So a reasonable use case is that someone might want either of the two Twiss parameters? ... which would mean it would make sense to have a `ParticleBeam.geomeotric_beta_x` and a `ParticleBeam.canonical_beta_x`?

**Another question:** What does Ocelot do? It seems to only have geometric Twiss parameters?

CHANGELOG.md
```diff
@@ -24,7 +24,7 @@
 - Add a Binder and link to `cheetah-demos` (see #482) (@smartsammler, @jank324)
 - PyTorch is now configured to use only deterministic algorithms during tests, preventing intermittent test failures (see #480) (@Hespe)
 - Make README example more copy-paste friendly, and generally improve it and the simple intro notebook in the docs. (see #493, #496) (@jank324, @ax3l)
-- Temporarily fix Ocelot test dependency to old 24.x.x commit where tests still pass (see #519) (@jank324)
+- Fix comparison tests to work with new PyPI release of Ocelot (see #513, #519) (@jank324, @cr-xu, @Hespe)
```
jank324, line 27:
```suggestion
- Fix comparison tests to work with new PyPI release of Ocelot. Remove Twiss tests where they are not needed. Increase tolerances where Cheetah and Ocelot follow slightly different models. (see #513, #519) (@jank324, @cr-xu, @Hespe)
```

Hespe (approved):

Issue #523, opened 2025-07-25
R56 sign of dipole magnet + feature addition


seongyeolk:
Hi all,

Hello, this is Seongyeol Kim of PAL-XFEL in Korea. 
I would like to propose a fix of the dipole element for bunch length calculation and addition of the feature for GPSR projects.

-- In case of sign for terms related to R56:
When I tried to do the benchmarking with Bmad-X and elegant simulation, I found that the bunch length does not match with the elegant simulation result after the bunch compressor. And I noticed that the sign of the R56 is following the Bmad-X's (line 352 of dipole.py), where the head-tail convention is flipped compared to the Cheetah and elegant. When I changed the sign, I checked that the tau distribution follows the elegant result.

-- Feature addition
I am currently working for the GPSR with Cheetah simulation (previously I worked with Bmad-X code), and I would like to add the feature to get the centroid shift of the beam with respect to the beam energy and set magnetic field. In the Bmad-X simulation, that feature is available by defining the reference energy of the beam in the dipole element (https://github.com/bmad-sim/Bmad-X/blob/main/bmadx/tracking_routines/track_a_bend.py), so it would be very good to add this feature in the Cheetah simulation. Then it will be straightforward to make the situation similar to the experimental setting where the applied current to the dipole magnet is always fixed but beam energy changes due to different cavity phase so that we could see the shift of the beam at the screen. (For cavity-based 6D GPSR, please refer to this paper: https://arxiv.org/abs/2502.20654)

Please feel free to let me know if there are any other things I should do for pull request. 
I am looking forward to hearing from you! 
Seongyeol

PS)
I worked on these for the Cheetah-based GPSR, and I modified "dipole.py" script in the branched version of my forked repo (https://github.com/seongyeolk/cheetah/tree/fix-dipolemagnet), so I would like to have a discussion with you for pull request.

jank324:
Hi Seongyeol,

those changes sound very interesting! Will you create a PR for them (ideally two separate PRs, one for the fix, another one for the features)? Then we would review and merge them.

seongyeolk:
> Hey, those changes sound very interesting! Will you create a PR for them (ideally two separate PRs, one for the fix, another one for the features)? Then we would review and merge them.

Sounds good! I will work on it and create separated requests.

PR #524, opened 2025-07-28
Fix sign of z_f related to R56 in dipole.py

seongyeolk:
## Description

Fix a sign of z_f in dipole magnet, which is related to R56

## Motivation and Context

As I reported the issue previously with #523, when I performed the benchmarking simulations using Cheetah and elegant, I found that the bunch length evolution after the bunch compressor is different compared to the elegant simulation. I noticed in dipole.py that z_f (updated longitudinal coordinate) in line 352 has - sign regarding R56 term, which is still following the Bmad-X convention. 

Since the head-tail convention in Cheetah simulation is same as the elegant (and flipped to Bmad-X), I would like to propose a change of the sign of it for bunch length calculation.

## Types of changes

## Checklist

### DIFF
--- a/cheetah/accelerator/dipole.py
+++ b/cheetah/accelerator/dipole.py
@@ -349,7 +349,10 @@ class Dipole(Element):
         z_f = (
             z
             + (beta * self.length.unsqueeze(-1) / beta0.unsqueeze(-1))
-            - ((1 + pz) * Lp / px_norm)
+            + (
+                (1 + pz) * Lp / px_norm
+            )  # In this line, sign should be + since head-tail convention
+            # is different compared to BmadX. Confirmed with elegant simulaton.
         )
 
         return x_f, px_f, y_f, py, z_f, pz


seongyeolk:
When I checked the error, I noticed that it actually includes the pytest for comparison between bmadx. After some thinking, I noticed that I was wrong about the bunch length calculation; since the Cheetah simulation does support the bmadx calculation, then during this procedure, the cheetah converts the z-coordinate information to bmadx convention, then it calculates the z_f under bmadx system. I should have thought about this. After this, I checked my calculation again, and I noticed that there was an issue with my calculation; I considered only the half of bunch compressor, so I compared it with the different condition.

Here are the benchmarking simulation between Cheetah and elegant. I ran the simulation using the lattice mentioned in the paper (https://arxiv.org/abs/2502.20654), and the results are shown below (for simplicity all the quadrupole magnets and cavity are off). These are (z-x) distribution after the half of bunch compressor, and z in elegant was obtained by converting it from "dt" with multiplying beta*c. 

halfBC_ZX_cheetah_elegant
ShalfBC_ZX_elegant_cheetah

As you can see, the particle distributions are pretty well overlapped. 
Calculated sigz is 
- elegant = 0.591758 mm
- cheetah = 0.594818 mm
- differentce = 3.06 um 

So, I am sorry for confusion, and please feel free to ignore this pull request. But still I would like to pull the request for adding the bmadx feature that can track the beam not along the design trajectory, by defining the arbitrary energy in the dipole magnet.

Thank you!
Seongyeol

jank324:
I'm glad this turned out to be fine. Thank you for checking @seongyeolk!

PR #534, opened 2025-08-07
Methods for getting the element names / indices for segment

roussel-ryan:
## Description

Adds property that lists the element names as well as a method that returns element index for a given element name.

## Motivation and Context

## Types of changes

## Checklist

### DIFF
--- a/cheetah/accelerator/segment.py
+++ b/cheetah/accelerator/segment.py
@@ -60,16 +60,18 @@ class Segment(Element):
     @property
     def element_names(self) -> list[str]:
         """
-        List of names of the elements in the segment.
-        :return: List of names of the elements in the segment.
+        Ordered list of names of the elements in the segment.
+        
+        :return: Ordered list of names of the elements in the segment.
         """
         return [element.name for element in self.elements]
 
-    def get_element_index(self, element_name: str) -> int:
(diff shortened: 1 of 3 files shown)


roussel-ryan:
In a number of cases I need to access the element by index in the list of elements, for example, I want to get the element directly upstream of another element without knowing the name

cheetah/accelerator/segment.py
```diff
@@ -57,6 +57,25 @@ def __init__(
             else:
                 self.__dict__[element.name] = element
 
+    @property
+    def element_names(self) -> list[str]:
+        """
+        List of names of the elements in the segment.
+        :return: List of names of the elements in the segment.
```
jank324, line 64:
```suggestion
        Ordered list of names of the elements in the segment.
        
        :return: Ordered list of names of the elements in the segment.
```

cheetah/accelerator/segment.py
```diff
@@ -57,6 +57,25 @@ def __init__(
...
+        """
+        return [element.name for element in self.elements]
+
+    def get_element_index(self, element_name: str) -> int:
```
jank324, line 68:
```suggestion
    def element_index(self, element_name: str) -> int:
```

PR #538, opened 2025-08-08
Speed improvements (from profiling)

jank324:
## Description

This PR does various things to improve Cheetah's speed:
 - Remove `verify_device_and_dtype`, which comes with changes to how dtypes are handled by Cheetah in general. The changed variant is now more in line with PyTorch itself.
 - Reduce the number of `torch` operations, because each of them comes with a kernel invocation that is much more costly than the computation itself for scalars. We look for operations that do as much as possible with as few kernel invocations as possible.
 - Choose faster PyTorch operations.
 - Reduce branching (this also restores some breaks in the compute graphs). To do so, we just ignore some things like divisions by zero, when they would be unphysical, or we try to rearrange equations, such that branching edge cases no longer appear.

## Motivation and Context

From the last 0.5.x release to the first 0.6.x release, Cheetah slowed down by about 5-6x. It appears that the primary cause of this is the full use of tensors in all computations, even scalar ones. The latter is needed to maintain differentiability. An additional slowdown of about 30% was caused between 0.6.x releases and the most recent 0.7.5 release. The reasons for this are multifold, but include for example `verify_device_and_dtype`.

On a side note, operations like `torch.where` and `torch.clamp` can break the compute graph, by selecting a subgraph that no longer contains results from parameters that one may be optimising over.

## Types of changes

## Checklist

### DIFF
--- a/.github/workflows/format.yaml
+++ b/.github/workflows/format.yaml
@@ -30,4 +30,4 @@ jobs:
         uses: py-actions/flake8@v2
         with:
           max-line-length: "88"
-          plugins: "flake8-bugbear flake8-black"
+          plugins: "flake8-bugbear"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,18 +15,21 @@
 - Beam classes now account for dispersion. Dispersion correction is included in the Twiss and emittance computations. Dispersion arguments are added to `from_parameters` and `from_twiss` beam initialisation methods. (see #540) (@cr-xu)
 - Add convenience methods to `Segment` for getting an ordered list of all element names and the index of a specific element by its name (see #534) (@roussel-ryan, @jank324)
 - First- and second-order transfer maps are now cached resulting in potential speed-ups of up to 10x and more (see #532) (@jank324)
+- Methods for creating `ParticleBeam` instances from distributions via stochastic sampling now make sure that the statistics of the generated particles match the desired distribution (see #546) (@cr-xu)
 
 ### 🐛 Bug fixes
 
 - Shorten `__repr__` of `Segment` for large lattices to prevent debugging slowdowns (see #529) (@Hespe)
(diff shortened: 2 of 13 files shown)


cheetah/utils/kde.py
```diff
@@ -59,7 +59,7 @@ def _kde_marginal_pdf(
                 f"Weights must have the same shape as values. Got {weights.shape}"
             )
 
-    residuals = values - bins.repeat(*values.shape)
+    residuals = values - bins.expand(*values.shape)
```
jank324, line 62: Is the `expand` even needed here? Shouldn't it broadcast automatically?

cheetah/accelerator/vertical_corrector.py
```diff
@@ -38,34 +34,30 @@ def __init__(
...
         )
 
-        tm = torch.eye(7, device=device, dtype=dtype).repeat((*vector_shape, 1, 1))
-        tm[..., 0, 1] = self.length
-        tm[..., 2, 3] = self.length
-        tm[..., 3, 6] = self.angle
-        tm[..., 4, 5] = -self.length / beta**2 * igamma2
+        tm = torch.eye(7, **factory_kwargs).expand((*length.shape, 7, 7)).clone()
+        tm[..., (0, 2, 3, 4), (1, 3, 6, 5)] = torch.stack(
+            [length, length, angle, -length / beta * beta * igamma2], dim=-1
```
Hespe, line 59:
```suggestion
            [length, length, angle, -length / (beta * beta) * igamma2], dim=-1
```

jank324:
I just realised that there are places where we compute a reducing operation like `sum` or `mean`, and then `unsqueeze` again. It would be much more efficient to call the operation with `keep_dims=True`.

jank324:
benchmark_commits_plot

jank324:
benchmark_commits_plot

PR #546, opened 2025-08-22
Match moments when sampling `ParticleBeam` with `from_distribution`

cr-xu:
## Description

- Fix the issue that randomly sampled particles will not match exactly to the desired cov matrix. This also helps the `from_twiss` and `from_parameter` generation as they call the `from_distribution`.
- Added a general `match_distribution_moments` function in `utils/statistics.py` which can be later used to match other distributions beyond ParticleBeam initialization.

## Motivation and Context

## Types of changes

## Checklist


cr-xu:
I'm really puzzled by the ubuntu github runner failure. I can't seem to reproduce the error on linux and macos.

image

Hespe:
I don't have a solution right now, but interestingly I can actually reproduce the issue on Ubuntu 20.04 with Python 3.13 with the same three tests failing. Not so much on Windows though.

jank324:
I cannot reproduce it on macOS.

Hespe:
We observed something interesting: `MultivariateNormal.rsample` returns slightly different samples on Unix and Windows systems if run with `torch.float32` even with the same random seed.

jank324:
> > I think this can be merged now, but I'm not sure I really understand what this does that `transformed_to` doesn't do.
> 
> `transformed_to` only respects the diagonal terms, ideally it should also be replaced with `match_moments` which scales & rotates the whole distribution. (something like `transformed_to(new_mu, new_cov)`)

Okay. @cr-xu, did you check the comment by Copilot? After that this should be merged.

cr-xu:
@jank324  What the heck, the Pipeline fails at the most random place, a TCL/TK error by matplotlib?

PR #576, opened 2025-09-28
Add 2D space charge element

austin-hoover:
## Description

This PR adds a `SpaceChargeKick2D` element. The code is pretty much the same as the current `SpaceChargeKick` element. The integrated 2D Green function is taken from https://www.sciencedirect.com/science/article/abs/pii/S0021999104000282?via%3Dihub. 

I have also renamed `SpaceChargeKick` to `SpaceChargeKick3D`.

## Motivation and Context

2D space charge is valid for long bunches without variation along the z axis.

## Types of changes

## Checklist


austin-hoover:
I've added a benchmark against the KV envelope equations in a drift and FODO lattice: https://github.com/austin-hoover/cheetah/tree/sc2d/tests/test_space_charge_kick_2d. Temporarily putting this in `/tests` directory.

fig_rms_beam_sizes

fig_rms_beam_sizes

cheetah/accelerator/space_charge_kick.py
```diff
@@ -691,3 +691,572 @@ def defining_features(self) -> list[str]:
...
+        self.register_buffer_or_parameter(
+            "grid_extent_y",
+            (
+                grid_extent_y
+                if grid_extent_y is not None
+                else torch.tensor(3.0, **factory_kwargs)
+            ),
+        )
+
+    def _deposit_charge_on_grid(
```
roussel-ryan, line 765: could you write general functions for depositing charge on 1/2/3 D grids and put it in a separate? This is something that we want to do for wakefields, images as well

cr-xu: Not yet. It would be great if you can add this feature.

aswaroop99: Hey Austin, Ryan mentioned this to me in our last meeting. I have code for 1d and 2d grids from our wakefields work which I can send over. I think it may be a good idea to start with 3 separate charge deposition functions, as the 1d case is especially simplified.

austin-hoover: Great! Want to start a new PR with your work?

RemiLehe: I think it would make sense to first finalize/merge this PR, and then later refactor the code to expose a general grid deposition function that could be used in other parts of the code (incl. in the code that generates beam images).

RemiLehe:
In addition, I think it would be good to have tests for the `SpaceChargeKick2D` that run as part of the CI.
This could be either the new tests that you added (if it is fast enough), or another test inspired e.g. by the existing CI tests for `SpaceChargeKick3D`. 

A key aspect of the CI tests is that they need automated checks (typically with `assert ` statements that result in a Python exception whenever the test criterion is not met: see e.g. https://github.com/desy-ml/cheetah/blob/master/tests/test_space_charge_kick.py#L126). 
My understanding is that the tests that you added currently produces plots, and require a human to look at them and check agreement, but do not yet have the automated `assert` statements. (Please correct me if I overlooked something here.)

austin-hoover:
We could compute derivative of single particle $x'$ with respect to $s$?

$$
x'' = \frac{dx'}{ds} = \frac{2Q}{(r_x + r_y)} \frac{x}{r_x}.
$$

tests/test_vectorized.py
```diff
@@ -347,7 +347,8 @@ def test_vectorized_screen_2d(BeamClass, method):
             cheetah.Marker,
             cheetah.Screen,
             cheetah.Segment,
-            cheetah.SpaceChargeKick,
+            cheetah.SpaceChargeKick3D,
+            cheetah.SpaceChargeKick2D,
```
jank324:
```suggestion
            cheetah.SpaceChargeKick2D,
            cheetah.SpaceChargeKick3D,
```

jank324 (changes requested):

PR #588, opened 2025-11-04
Update ARES lattice JSON

cr-xu:
## Description

- Update ARES lattice json to new standards in the document example
- Update the gradient-based optimization notebook to use non-zero initialization for magnet strengths

## Motivation and Context

## Types of changes

## Checklist


jank324:
I remove the change in `gradientbased.ipynb` because starting from non-zero is part of the story line in that notebook, hence it did not make sense to do it in the first "bad" example.

Nevertheless, currently that first bad example is more broken than it should be because of bad gradients. This should be fixed by #553.

The notebook was also broken because of the old state of the LatticeJSON file.

Given that those are two different problems, I think we should merge this as a LatticeJSON fix (current HEAD), and open an issue for the other issue that will then be fixed with #553.

@cr-xu, do you agree?

jank324 (approved):

PR #591, opened 2025-11-13
Superimposed element

roussel-ryan:
## Description

This PR implements a `Superimposed` element class that supports placing an additional element in the center of another element.

The `Superimposed` class has a `base_element` and `superimposed_element`. During tracking the `base_element` is split in half and the `superimposed_element` is placed in between both halves at runtime. This allows modification of the base / superimposed element properties in place, without worrying about correctly accounting for the eventual splitting of the element. 

## Motivation and Context

This PR addresses problems with setting element attributes when trying to model beamlines that have superimposed elements. For example, previous cheetah lattices split the elements in the lattice itself, meaning that the length and strength elements were not equal to the full element properties. This created unexpected bugs when setting element parameters (e.g. geometric focusing strength) that depend on the length of the element (for example, when converting from integrated gradient to

Addresses #592 

## Types of changes

## Checklist


phys-cgarnier:
@cr-xu Hi, can I have the workflows approved?

phys-cgarnier:
@cr-xu What should I do about the pkl file used for consistency test?

phys-cgarnier:
@cr-xu Also, I don't know that it makes sense to implement splitting for the super imposed element unless splitting is grabbing the already split components in the superimposed element.

cr-xu:
> @cr-xu What should I do about the pkl file used for consistency test?

For a new element being implemented, you could just generate those and add those to the tests. They are mostly for keeping track that we don't accidentally introduce breaking changes.

cr-xu:
> @cr-xu Also, I don't know that it makes sense to implement splitting for the super imposed element unless splitting is grabbing the already split components in the superimposed element.

That probably depends on the way how it is intended to be used. From the current code I see that the superimposed element is put at the center of the other element. So if the real element is not centered, we would already split the base element accordingly and only construct the SuperImposed with split elements, right? In that case, the splitting doesn't make sense and you can just leave it empty.

phys-cgarnier:
@cr-xu 
I cannot set the PR to ready for review, would you mind doing this?

cheetah/accelerator/__init__.py
```diff
@@ -18,3 +18,4 @@
 from .transverse_deflecting_cavity import TransverseDeflectingCavity  # noqa: F401
 from .undulator import Undulator  # noqa: F401
 from .vertical_corrector import VerticalCorrector  # noqa: F401
+from .superimposed import SuperimposedElement  # noqa: F401
```
jank324: We usually sort imports alphabetically. Running `isort` should easily make sure all imports are sorted and arranged correctly and consistently.

cheetah/accelerator/segment.py
```diff
@@ -136,6 +136,8 @@ def flattened(self) -> "Segment":
         for element in self.elements:
             if isinstance(element, Segment):
                 flattened_elements += element.flattened().elements
+            elif hasattr(element, "subelements"): # e.g. SuperimposedElement
+                flattened_elements += element.subelements
             else:
```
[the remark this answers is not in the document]
jank324: This comment seems reasonable to me.

cheetah/accelerator/superimposed.py
```diff
@@ -0,0 +1,152 @@
...
+        Parameters
+        ----------
+        base_element : Element
+            The base element over which other elements are superimposed.
+        superimposed_element : Element
+            Segment of elements to be superimposed at the center of base element
+        name : str, optional
+            The name of the segment. If None, a default name is generated.
+        sanitize_name : bool, optional
+            Whether to sanitize the name to ensure it is valid.
```
jank324: This does not match the doctoring style we use in Cheetah.

cheetah/accelerator/superimposed.py
```diff
@@ -0,0 +1,152 @@
...
+    ie. where one element is placed over another at the center of the base element.
+
+    """
+
+    def __init__(
+        self,
+        base_element: Element,
+        superimposed_element: Segment,
+        name=None,
+        sanitize_name=False,
```
jank324: Type annotations are incomplete.

cheetah/accelerator/superimposed.py
```diff
@@ -0,0 +1,152 @@
...
+        elif isinstance(superimposed_element, Element):
+            self.superimposed_element = Segment(
+                elements=[superimposed_element],
+                name=f"{superimposed_element.name}_segment",
+            )
+        else:
+            raise TypeError(
+                f"superimposed_element must be a Segment or Element subclass, "
+                f"got {type(superimposed_element).__name__}"
+            )
```
jank324: This is a bit risky, because `isinstance(some_segment, Element) == True`.

cheetah/accelerator/superimposed.py
```diff
@@ -0,0 +1,152 @@
...
+
+    # self._buffers["length"] = self.base_element._buffers["length"]
```
jank324: Dead code should not be left in here.

cheetah/utils/cache.py
```diff
@@ -14,6 +14,7 @@ def cache_transfer_map(func):
     @functools.wraps(func)
     def wrapper(self: Element, energy: torch.Tensor, species: Species) -> torch.Tensor:
         # Caching is not supported if any of input tensors require gradients
+
```
jank324:
```suggestion
```

cheetah/utils/cache.py
```diff
@@ -59,10 +61,11 @@ def wrapper(self: Element, energy: torch.Tensor, species: Species) -> torch.Tens
             cache["result"] = func(self, energy, species)
 
             cache["feature_validity_key"] = feature_validity_key
+            cache["feature_tensors"] = feature_tensors
             cache["energy"] = energy.clone()
             cache["num_elementary_charges"] = species.num_elementary_charges.clone()
             cache["mass_eV"] = species.mass_eV.clone()
-
+        # returns after
```
jank324:
```suggestion

```

cheetah/utils/cache.py
```diff
@@ -59,10 +61,10 @@ def wrapper(self: Element, energy: torch.Tensor, species: Species) -> torch.Tens
             cache["result"] = func(self, energy, species)
 
             cache["feature_validity_key"] = feature_validity_key
+            cache["feature_tensors"] = feature_tensors
             cache["energy"] = energy.clone()
             cache["num_elementary_charges"] = species.num_elementary_charges.clone()
             cache["mass_eV"] = species.mass_eV.clone()
```
jank324, line 64:
```suggestion
            cache["mass_eV"] = species.mass_eV.clone()
            
```

cheetah/utils/cache.py
```diff
@@ -59,10 +61,11 @@ def wrapper(self: Element, energy: torch.Tensor, species: Species) -> torch.Tens
...
-
+
```
jank324:
```suggestion
        
```

cheetah/utils/cache.py
```diff
@@ -59,10 +61,11 @@ def wrapper(self: Element, energy: torch.Tensor, species: Species) -> torch.Tens
...
```
jank324:
```suggestion

```

cheetah/latticejson.py
```diff
@@ -59,9 +71,9 @@ def convert_segment(segment: "cheetah.Segment") -> tuple[dict, dict]:
 
             elements.update(segment_elements)
             lattices.update(segment_lattices)
+
```
jank324:
```suggestion
```

jank324:
Just for my understanding: Is the superimposed element basically a `Segment` with `[base -> superimposed -> base]`? What happens if `superimposed` has a length?

roussel-ryan:
For now we should restrict superimposed elements to those who have zero length

cheetah/accelerator/superimposed.py
phys-cgarnier: good catch!
I made it more explicit that the type should be a subclass of element, but, not element and not segment.

cheetah/accelerator/superimposed.py
```diff
@@ -0,0 +1,152 @@
...
+        return all([el.is_skippable for el in self.subelements])
+
+    def plot(
+        self, s: float, vector_idx: tuple | None = None, ax: plt.Axes | None = None
+    ) -> plt.Axes:
+        return self.base_element.plot(s, vector_idx=vector_idx, ax=ax)
+
+    @property
+    def defining_features(self) -> list[str]:
+        return super().defining_features + ["base_element", "superimposed_element"]
```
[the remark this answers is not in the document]
phys-cgarnier: I verified the current clone behavior by checking object IDs and tensor data_ptr() values after cloning. The top-level object, nested elements, and tensor storage are all independent, so I don’t see an actual memory-sharing issue here.

cheetah/accelerator/superimposed.py
```diff
@@ -0,0 +1,152 @@
...
+
+        if name == "length" and hasattr(self, "base_element"):
+            base = getattr(self, "base_element", None)
+            if base is not None:
+                base.length = self.length
+
+    def track(self, incoming: Beam) -> Beam:
+        if self.is_skippable:
+            return self._track_first_order(incoming)
+        else:
```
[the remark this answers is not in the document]
phys-cgarnier: Copilot is saying that is when all elements are skippable, tracking the beam like it is passing through a drift of self.length which is defaulted to zero. So if my setattr for length never gets called then the self.length != self.base_element.length 

I will add 
self.length = self.base_element.length.clone()
after init. And redo the test_consistency generation.

cheetah/accelerator/superimposed.py
```diff
@@ -0,0 +1,152 @@
...
+        )
+        second = type(base)(
+            length=half_length,
+            name=f"{base.name}#1",
+            sanitize_name=False,
+            dtype=base.length.dtype,
+            device=base.length.device,
+            **kwargs,
+        )
+
```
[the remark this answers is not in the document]
phys-cgarnier: base element should aways have length but I suppose I can make an explicit check.

cheetah/accelerator/superimposed.py
```diff
@@ -0,0 +1,152 @@
...
+from cheetah.accelerator import Segment
+from cheetah.accelerator.element import Element
+from cheetah.particles.beam import Beam
+from cheetah.particles.species import Species
+from cheetah.utils.unique_name_generator import UniqueNameGenerator
+
+generate_unique_name = UniqueNameGenerator(prefix="unnamed_element")
+
+
```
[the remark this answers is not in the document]
phys-cgarnier: Should I remove this? I don't think its necessary but other elements have there own copies of it.

cheetah/accelerator/segment.py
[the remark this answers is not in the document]
phys-cgarnier: Working on a this

tests/test_superimposed.py
```diff
@@ -0,0 +1,135 @@
...
+    segment_dict = {"elements": elements, "lattices": lattices}
+
+    # test conversion back to segment
+    reconstructed_segment = parse_segment("FullSegment", segment_dict)
+    assert (
+        reconstructed_segment.SuperimposedBPM.base_element.k1
+        == superimposed_element.base_element.k1
+    )
+
+    full_segment.to_lattice_json("full_segment.json")
```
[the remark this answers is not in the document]
phys-cgarnier: Set this to dump to a tmp file and check the path exists after calling the function. full_segment.json should no longer be written into the directory.

cheetah/accelerator/superimposed.py
```diff
@@ -0,0 +1,152 @@
...
+        name=None,
+        sanitize_name=False,
+    ):
+        """
+        Parameters
+        ----------
+        base_element : Element
+            The base element over which other elements are superimposed.
+        superimposed_element : Element
+            Segment of elements to be superimposed at the center of base element
```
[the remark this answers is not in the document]
phys-cgarnier: updated type annotations

tests/test_superimposed.py
```diff
@@ -0,0 +1,135 @@
...
+    # set the quadrupole strength through the superimposed segment
+    # and ensure it propagates
+    superimposed_segment.base_element.k1 = torch.tensor(3.0)
+    quad.k1 = torch.tensor(3.0)
+
+    tm_expected2 = quad.first_order_transfer_map(energy, species)
+    tm = superimposed_segment.first_order_transfer_map(energy, species)
+    #tm_expected2 = quad.first_order_transfer_map(energy, species) 
+    # bug where tm is stale if the cache is not properly invalidated when the quadrupole strength is changed
+    assert not torch.allclose(tm_expected, tm_expected2)  # should not be close to the old transfer map
```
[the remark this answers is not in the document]
phys-cgarnier: removed these comments breaking line length linting.

cheetah/accelerator/__init__.py
phys-cgarnier: ran isort.

cheetah/accelerator/superimposed.py
phys-cgarnier: matched the style to what I observed in quadrupole.py

cheetah/accelerator/superimposed.py
phys-cgarnier: updated

phys-cgarnier:
@jank324 
I am confused why pytest 3.12 fails the test_vectorized_indexing with a quad.
I haven't pushed any code that could affect the quads since I modified the cache_transfer_map property. I actually cannot recreate this test locally, although my OS is macos

The values shown appear quite close actually, maybe the arg equal_nan fixes this. 

` assert torch.allclose(sub_beam.particles, outgoing.particles[:3], equal_nan = True)`

cheetah/accelerator/superimposed.py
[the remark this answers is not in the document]
phys-cgarnier: Done

cheetah/accelerator/segment.py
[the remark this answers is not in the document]
jank324: Actually, is it the desired behaviour, that a `Superimposed` element is dissolved when `Segment` hierarchies are flattened?

roussel-ryan:
Flattening the superimposed element should decompose it into a un nested state IMO

cheetah/accelerator/superimposed.py
```diff
@@ -0,0 +1,154 @@
...
+        base = self.base_element
+
+        kwargs = {
+            feature: getattr(base, feature)
+            for feature in base.defining_features
+            if feature != "length" and feature != "name"
+        }
+
+        first = type(base)(
+            length=half_length,
```
jank324: Just creating an element with half the length isn't the physically correct way to split an element for all elements. That's why we have the `.split()` method. But note that that method isn't implemented by all elements classes.

cheetah/accelerator/superimposed.py
```diff
@@ -0,0 +1,158 @@
...
+        factory_kwargs = {"device": device, "dtype": dtype}
+        super().__init__(name=name, sanitize_name=sanitize_name, **factory_kwargs)
+
+        self.base_element = base_element.to(**factory_kwargs)
+        if type(superimposed_element) is Segment:
+            self.superimposed_element = superimposed_element.to(**factory_kwargs)
+        elif type(superimposed_element) is not Element and isinstance(
+            superimposed_element, Element
+        ):
+            self.superimposed_element = Segment(
```
jank324: I'm not sure about this implementation. I think a better way to do it would be to take `base_element`, split it in two with `.split()` and then create a private `_segment` property that is actually a `Segment` and hold the three subelements in there. Things like transfer map computation could then just forward to the respective method of `_segment`. One would probably have to keep a hold of the base element and redo the split when `.length` or any other property is changed by the user.

Doing it this way would make the code for this class a lot more leaner and mean it automatically inherits features and improvements from `Segment` without having to be a subclass of it. Using `.split()` would also fix the fact the current implementation may be physically incorrect for some elements.

Does that sound like it makes sense?

roussel-ryan: this could be better, although I'm particularly concerned about making sure that we can directly modify the base element while making sure that **any** change would be propagated to the slit elements

cheetah/accelerator/superimposed.py
roussel-ryan: For now, I only want to consider the case were the superimposed element is at the halfway point. Anything beyond that requires a lot more code for arbitrary splits. As such, I want to leave that to a future PR

roussel-ryan: So I misunderstood @jank324 's comment, we should use split with a resolution of half the length

cheetah/accelerator/superimposed.py
phys-cgarnier: @jank324 How do you recommend syncing superimposed.base_element.length to superimposed.length, what would you do to updated superimposed.length on write to superimposed.base_element.length

phys-cgarnier:
@jank324 
I made the requested changes, and all tests appear to be passing locally.

jank324:
Because it was revealed and a minor issue for this PR, the following fix is made as part of this PR (see most recent commit): Fix an issue in consistency tests where some elements were incorrectly not tested with `ParameterBeam`

jank324 (approved):

cheetah/accelerator/superimposed.py
```diff
@@ -0,0 +1,92 @@
...
+
+generate_unique_name = UniqueNameGenerator(prefix="unnamed_element")
+
+
+class Superimposed(Element):
+    """
+    A segment that represents a superimposed structure in an accelerator, i.e. where one
+    element is placed over another at the center of the base element.
+
+    NOTE: Chaning either `base_element` or `superimposed_element` after initialisation
```
Hespe:
```suggestion
    NOTE: Changing either `base_element` or `superimposed_element` after initialisation
```

cheetah/accelerator/superimposed.py
```diff
@@ -0,0 +1,92 @@
...
+        will lead to unexpected behaviour. If you need to change either of these
+        elements, please create a new instance of `Superimposed`.
+
+    :param base_element: The base element at the center of which the superimposed
+        element is placed.
+    :param superimposed_element: Element to be placed at the center of the base element.
```
Hespe, line 24: Should we document the zero-length requirement?

cheetah/accelerator/superimposed.py
```diff
@@ -0,0 +1,92 @@
...
+
+        self.base_element = base_element
+        self.superimposed_element = superimposed_element
+
+        base_element_halves = base_element.split(base_element.length / 2.0)
+        self.elements = [
+            base_element_halves[0],
+            superimposed_element,
+            base_element_halves[1],
+        ]
```
Hespe: Should `self.elements` be a property that refers to the elements of the internal segment?

Hespe: Or rather be removed completely?

cheetah/accelerator/superimposed.py
```diff
@@ -0,0 +1,92 @@
...
+    ) -> torch.Tensor:
+        return self._segment.first_order_transfer_map(energy, species)
+
+    def track(self, incoming: Beam) -> Beam:
+        return self._segment.track(incoming)
+
+    def plot(
+        self, s: float, vector_idx: tuple | None = None, ax: plt.Axes | None = None
+    ) -> plt.Axes:
+        return self._segment.plot(s, vector_idx=vector_idx, ax=ax)
```
Hespe, line 84: Maybe we should have a custom plot function? The implementation of `Segment` will show a split in the `base_element`. Which is of course fine, so this is only cosmetic.

tests/test_elements.py
```diff
@@ -343,6 +340,24 @@ def test_consistency(element, beam_cls):
         if not isinstance(incoming_beam, beam_cls):
             incoming_beam = incoming_beam.as_parameter_beam()
 
+    try:
+        actual_outgoing_beam = element.track(incoming_beam)
+    except Exception as e:
```
Hespe: I would vouch for using a more specific Exception Class here

tests/test_vectorized.py
```diff
@@ -417,7 +418,10 @@ def test_vectorized_parameter_beam_creation():
     "ElementClass", [cheetah.HorizontalCorrector, cheetah.VerticalCorrector]
 )
 def test_broadcasting_corrector_angles(ElementClass):
-    """Test that broadcasting rules are correctly applied to with corrector angles."""
+    """
+    Test that broadcasting rules are correctly applied to elements with corrector
+    angles.
+    """
```
Hespe:
```suggestion
    """Test that broadcasting rules are correctly applied to corrector angles."""
```

Hespe (changes requested):

cheetah/accelerator/superimposed.py
jank324: It was dead code actually. I had forgotten to remove it.

cheetah/accelerator/superimposed.py
jank324: I think that's fine for a later date if ever needed. It would show a split in the base element, but with zero distance in between. So the way plotting currently works, it would look like a base element, with the superimposed element right in the middle, which I think is very representative.

tests/test_elements.py
jank324: I switched it to `AssertionError`

Hespe (approved):

Issue #543, opened 2025-08-21
Write common utility function for charge deposition on a grid


RemiLehe:
Differentiable charge deposition on grid (with first order spline, a.k.a. Cloud-In-Cell) can be used in several parts of Cheetah:
- in space-charge (with a 3D grid)
- in wakefield calculation (with a 1D grid for longitudinal wakefield, or a 2D grid for transverse wakefield)
- in computing beam images

I would be good to have a unique function to do all of these.

## Potential signature:

```
deposit_charge( x, y, z, weight, survival_probability, grid, dx, dy, dz, xmin, ymin, zmin ):
```
`x`, `y`, `z`, `weight`, `survival_probability`: 1D arrays with one element per particle (or as an alternative: `positions`, which would be a `(batch_size, N_particles, d)` array, where `d` is the same dimensionality as the grid.
`grid`: array of size `(batch_size, nx, ny, nz)` (in 3D)
dx, dy, dz, xmin, ymin, zmin : scalar metadata for the grid. (or as an altenative, `cell_size` and `lower_bound`: array of shape `(batch_size, d,)` )

The function would modify the array `grid` in place.

## Some questions:

- Should we use `scatter_add_` or `index_put_`: we need to understand the differences and implication for performance.

cr-xu:
We can even split the method into `deposit_charge_1d`, `deposit_charge_2d`, `deposit_charge_3d` as the current function has many specific lines for 3D  cases anyway.

cr-xu:
This is implemented now in the #625 , with a new screen method.

Once that PR is merged, we can work on refactoring the spacechargekick to use the general method, and start thinking about a wake element

PR #625, opened 2026-03-16
Feature: Screen Charge Deposition

### LINKED ISSUE #433: Use kernel with compact support in screen KDE
When using the screens with KDE, the distribution of beam particles is convolved with a Gaussian kernel. 

Since a Gaussian kernel has infinitely wide support, every single beam particle contributes to every single pixel, resulting in `O( N_beam x N_pixels )` operations, which can be quite expensive.

As an alternative, we could use a kernel that has compact kernel, i.e. a kernel for which each particle affects only its neighboring pixels. One example is the so-called Cloud-In-Cell kernel that is already used in the (differentiable) space-charge implementation in Cheetah: 
https://github.com/desy-ml/cheetah/blob/master/cheetah/accelerator/space_charge_kick.py#L88

cr-xu:
## Description

- Add a general Cloud-in-Cell (CIC) charge deposition implementation for an arbitrary number of dimensions.
- Add `cloud-in-cell` method for Screen, that is differentiable and faster than the kde method.

Fixes #433 

Also partially addresses #543 

This PR merges @roussel-ryan 's fork https://github.com/roussel-ryan/cheetah/tree/diag0 into the master.

## Motivation and Context

## Types of changes

## Checklist


cheetah/accelerator/screen.py
```diff
@@ -24,8 +29,8 @@ class Screen(Element):
     :param misalignment: Misalignment of the screen in meters given as a Tensor
         `(x, y)`.
     :param method: Method used to generate the screen's reading. Can be either
-        "histogram" or "kde", defaults to "histogram". KDE will be slower but allows
-        backward differentiation.
+        "histogram", "kde", or "charge_deposition", defaults to "histogram".
+        KDE and charge_deposition methods allow backward differentiation.
```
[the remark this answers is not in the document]
jank324: This looks like an argument to make `"charge-in-cell"` the new default. It is very nearly as fast as `"histogram"`, while `"histogram"` leads to unexpected results because it breaks differentiability and vectorisation. It's better if those issues only happen when a user really needs that last little bit of performance.

Screenshot 2026-04-23 at 11 46 16

jank324:
I just discovered a fairly significant issue with the implementation as is: Cheetah assumes that there can be any number of vector dimensions, and how many there are is unknown. However, as it stands, the Cloud-in-Cell implementation assumes exactly one vector dimensions, and, I believe, relies on Tensor's number of dimensions to determine how many histogram dimensions are needed. This is fundamentally incompatible with how Cheetah does it. Instead, the Cloud-in-Cell implementation should follow the interface of [`numpy.histogramdd`](https://numpy.org/doc/stable/reference/generated/numpy.histogramdd.html).

roussel-ryan:
Are you sure about this? We've been doing arbitrary batch dimensions for gpsr using similar code.

jank324:
Well ... I just started checking how we might fix this. And I think it might actually be vector-dimension-agnostic, but the docstring and argument verification just say otherwise.

cheetah/utils/cloud_in_cell.py
```diff
@@ -0,0 +1,256 @@
...
+    bins: Sequence[torch.Tensor],
+    weights: torch.Tensor | None = None,
+) -> torch.Tensor:
+    """
+    Fast Cloud-in-Cell (CIC) charge deposition in 1D, 2D or 3D.
+
+    :param positions: List or tuple of particle position tensors, each of shape
+        `(..., N)`, leading with optional vector dimension. `N` is the number of
+        particles. The length `d` of `positions` determines the dimensionality
+        (1D, 2D or 3D).
```
jank324: I think this doesn't actually match what the function really does.

roussel-ryan: I agree

roussel-ryan: Wait, let me look at it a little more in detail

roussel-ryan: Wait, let me look at it a little more in detail

jank324: I changed it now to something that I believe reflects much better how it works (or should work). Interestingly, this is exactly backwards from what `np.histogramdd` does. I don't know if that's bad or irrelevant. I find the way we do it now intuitive as well.

roussel-ryan:
Hey Jan, we need to be very careful about not negativity impacting performance of charge deposition on GPUs when replacing the for loops with vectorization. I believe that some of the speed up may be due to memory allocation on GPUs. I would recommend not changing the algorithm until we have studied this more in detail.

jank324:
> Hey Jan, we need to be very careful about not negativity impacting performance of charge deposition on GPUs when replacing the for loops with vectorization. I believe that some of the speed up may be due to memory allocation on GPUs. I would recommend not changing the algorithm until we have studied this more in detail.

How exactly do you think the loops result in memory allocation speed ups?

I guess you are right (?) I've benchmarked the tensorised and looping implementations against each other (on CPU, MPS and CUDA). The looping implementation seems faster, especially with large and/or vectorised beams. For small beams the difference is negligible, for large beams it's multiple times faster.

I find that counterintuitive. Properly tensorised operations have in my experience been a lot faster than looping.

All of this suggests that memory is in fact the key factor here, but I don't quite understand how. The tensorised implementation has higher memory requirements (in theory it trades memory for speed), but by my estimate, it shouldn't need more than a couple GB even for a 100,000 particle beam vectorised by 100. I tested on an M1 Pro with 32 GB and a H100 with 96 GB, so this should not be the reason there are slow downs.

What is more, the number of memory allocations should also be lower in the tensorised implementation.

It could be something to do with caching ... but I don't really understand how that should be so different between both implementations.

And here comes the really confusing thing: I benchmarked the different components/sections of both implementations against one another. The results are ... inconclusive. Depending on the order in which I test them, the tensorised implementation is either a bit slower (between 20% and 4x-ish) or a lot faster (10x and more). As far as I can tell the one I run first seems to be the faster one. (Another pointer towards memory. But why? And what does that mean for the function as a whole?)

As for this PR: I am very intrigued by these results and I kind of want to understand what's going on now. I will play around a bit more, maybe document all the benchmarking results here properly, and then revert it back to the looping implementation, unless I find clear indications that (parts of) the tensorised one is/are faster.

roussel-ryan:
So this is one of the first things that I tried vibe coding with, here is the conversation I had with chatgpt on the topic, where I asked it to explain why we got performance improvements on GPU https://chatgpt.com/share/69fdeefb-e38c-83ea-ae89-c55121acb6d8 . I also got an interesting explanation when plugging in your comments above. My hope would be that we can merge in the original version of the code that we knows works well (so that we don't have to constantly work off of a branch of cheetah) and then add a benchmarking study / aim for additional improvements later on.

cr-xu:
> And here comes the really confusing thing: I benchmarked the different components/sections of both implementations against one another. The results are ... inconclusive. Depending on the order in which I test them, the tensorised implementation is either a bit slower (between 20% and 4x-ish) or a lot faster (10x and more). As far as I can tell the one I run first seems to be the faster one. (Another pointer towards memory. But why? And what does that mean for the function as a whole?)

This sounds like a memory issue, probably the benchmarks should be run independently / explicitly del objects after running each part?

jank324:
So after some tinkering, I figured out the issues. They were mostly to do with missing synchronisation and the fact that I randomly generated one of the inputs to produce a worst-case scenario for the GPU that completely disabled parallelisation.

Now the tensorised implementation is up to 2x faster than the looping implementation on all three devices.

I will clean this up, push it and then we can check that everything is working as it should.

jank324:
> So this is one of the first things that I tried vibe coding with, here is the conversation I had with chatgpt on the topic, where I asked it to explain why we got performance improvements on GPU https://chatgpt.com/share/69fdeefb-e38c-83ea-ae89-c55121acb6d8 . I also got an interesting explanation when plugging in your comments above. My hope would be that we can merge in the original version of the code that we knows works well (so that we don't have to constantly work off of a branch of cheetah) and then add a benchmarking study / aim for additional improvements later on.

This was super interesting in terms of getting a sense for how well vibe coding works today! I think the final implementation is not very elegant, but it does work quite well with I think a few bugs here and there. Nevertheless, I couldn't reproduce some of the claims it made regarding the speeds of `scatter_add_` and `index_add_`. Some of these seem a little contradictory to me, and in my testing, I got the better performance out of `scatter_add_` than the previous implementation with `index_add_`.

docs/requirements.txt
```diff
@@ -2,3 +2,4 @@ furo
 ipykernel
 myst-parser
 nbsphinx
+pandas
```
jank324:
```suggestion

```

jank324:
@cr-xu, I can't request a review from you. Maybe you can also have a very quick look at this before we merge.

roussel-ryan:
Sorry to be the bearer of bad news, but I ran the "cloud-in-cell" method vs. the original "charge_deposition" for the 6D phase space reconstruction workflow on an A100 (NVIDIA A100-SXM4-40GB) and here are the results: "cloud-in-cell" took 76 sec for 200 epochs, and "charge_deposition" took 17. Both are implemented in this branch https://github.com/roussel-ryan/cheetah/tree/diag0

Edit: I was also able to reproduce this on my own GPU (NVIDIA GeForce RTX 2070 SUPER) with the gpsr 6D example training: cloud-in-cell: 52.8s / 100 epochs,   charge_deposition: 19.6s / 100 epochs

jank324:
> Sorry to be the bearer of bad news, but I ran the "cloud-in-cell" method vs. the original "charge_deposition" for the 6D phase space reconstruction workflow on an A100 and here are the results: "cloud-in-cell" took 76 sec for 200 epochs, and "charge_deposition" took 17. Both are implemented in this branch https://github.com/roussel-ryan/cheetah/tree/diag0

No worries. This PR is becoming an interesting learning experience.

Weirdly, your times don't line up at all with my benchmarks on an A100. But I only tested `Screen.reading` on its own and the `cloud_in_cell_charge_deposition` function on its own in isolation. I also found that the sizes of everything (vector dimensions, screen dimensions, number of particles) affect the performance of both implementations differently and significantly.

@roussel-ryan can you give me some numbers, so I can compare my benchmarks? How many particle are you tracking? What size is your screen in pixels? And what are your vector dimensions?

I will try to post my benchmark results in here tomorrow.

cr-xu:
I'll try to have a closer look over the weekends. 

I've been using the latest version of this PR branch for some tests. I can agree that my training on a A100 GPU has also slowed down compared to earlier versions (not by 5 times though, more like 1.5-2 times) Couldn't really ping to which specific commit was the cause.
In my case I'm traking 100,000 particles, commonly with a batch size of 10 settings per step.

roussel-ryan:
Im tracking about 100k particles for 20 images that have about 150x150 pixels each. The batch dimension in this case ends up being 20x1 or 2x10 depending on the way you slice the data. LMK if there are any other questions

roussel-ryan:
@jp-ga may also be able to comment

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ### 🚨 Breaking Changes
 
+- Change default `method` of `Screen` from `"histogram"` to `"cloud-in-cell"` (see #625) (@roussel-ryan, @cr-xu, @jank324)
+
```
jank324, line 8: Need to remember to move **both** changelog entries up to the new release.

jank324:
I ran a couple of tests and did a little bit more optimising. So here are the results from that. All benchmarks use the vectorisations size, number of particles and screen resolution from @roussel-ryan.

First, here is the original tensorised implementation vs. the original loop-based implementation. These suggest that the tensorised implementation was slower on most devices, but slightly faster specifically on an A100 (here called "Linux - CUDA").
Screenshot 2026-06-17 at 16 33 39

Based on this, I optimised the tensorised implementation with the following changes:

### Optimisation of `cheetah/utils/cloud_in_cell.py`
- **Avoided Multidimensional Broadcasting**: Replaced generic multi-dimensional tensor broadcasting (which created massive intermediate tensors of size `(..., num_particles, num_corners, num_hist_dims)`) with specialised paths for 1D, 2D, and 3D histograms.
- **Contiguous Unpacking**: Unpacked the stacked `positions` tensor along the coordinate dimension at the very beginning and called `.contiguous()` on the resulting sliced coordinates. This allows downstream operations (such as `.floor()`, `.clamp()`, and arithmetic) to execute on contiguous blocks of memory, unlocking CPU vectorisation and GPU cache locality.
- **Early Extent Filtering**: Masked out-of-bounds particles once at the beginning using a particle-level `in_extent` filter. This completely removed the need for expensive individual logical AND masks for each of the corners.

### Optimisation of `cheetah/accelerator/screen.py`
- **Cached Property Access**: Avoided calling the `pixel_bin_edges` property multiple times (which triggered `torch.linspace` 8 times per tracking call).
- **Direct GPU Allocation**: Created the `extent` tensor directly on the target device, avoiding redundant CPU-to-GPU memory copies during element tracking.

The resulting speed benchmark shows that now the tensorised implementation is consistently faster.
benchmark_comparison

The specialised paths for 1D, 2D and 3D result in a lot of extra code, so I checked if it gives enough of a speed up to justify that. The following plot shows the optimised general implementation vs. that with the specialised paths.
specialized_paths_comparison

And it turns out one of the most significant bottlenecks wasn't in the cloud-in-cell implementation itself, but rather in `Screen.reading`. The following plot shows the optimised cloud-in-cell implementation with and without the optimised implementation of `Screen.reading`.
screen_optimization_comparison

jank324:
And a comparison between the screen methods at this point
screen_methods_comparison

jank324:
@roussel-ryan can you point me to the code that produced those bad results for you? I wasn't able to reproduce the 4.5x slow down you saw with my own benchmark code.

roussel-ryan:
https://github.com/roussel-ryan/gpsr/blob/main/docs/examples/example_reconstruction_6d.ipynb

roussel-ryan:
and/or https://github.com/roussel-ryan/gpsr/blob/main/docs/examples/example_reconstruction_4d.ipynb

jank324:
chenran_comparison

roussel-ryan:
What are the number of particles being tracked and the pixel dimensions of the screens for these benchmarks?

jank324:
> What are the number of particles being tracked and the pixel dimensions of the screens for these benchmarks?

100k particles, 150x150 screen for (20, 1) and (2, 10) vectorisations, as you suggested in an earlier comment.

> Im tracking about 100k particles for 20 images that have about 150x150 pixels each. The batch dimension in this case ends up being 20x1 or 2x10 depending on the way you slice the data. LMK if there are any other questions

jank324:
I have also run a speed benchmark of the 6D GPSR with both versions of this PR.

I initially observed that the current tensor-based version took around 2x as long as the loop-based one. However, it turned out that this was only because it came first in the benchmark notebook. My best guess that this is because of how PyTorch handles memory allocation. The highly oversimplified summary is that PyTorch likes to do its own memory allocation. So when the user requests memory for a tensor, PyTorch will go and get that memory from the OS and then internally allocate it to that tensor. But later when the user destroys that tensor, PyTorch will hold on to that memory and use it the next time the users asks for memory. That second memory allocation not involving the OS is MUCH faster that the first. So whichever benchmark ran first in the kernel suffered from the expensive allocation of memory from the OS, while the second could get it very quickly from PyTorch internally. There could be other effects at play here as well, but this appears to be the best explanation to me.

So the benchmarks below include a kind of "warmup". Basically, I ran a warmup reconstruction for the first implementation, followed by a reconstruction that I actually timed, which is then followed by a warmup reconstruction for the second implementation and a timed reconstruction for the second implementation. In the plot these are referred to as "warmed up". With this warm up the order in which they are run should be irrelevant, but for completeness, the tensor-based implementation was run first and the loop-based second (on the same kernel).

As a sanity check, I also ran cold-start (no warmup) benchmarks in separate notebooks with separate kernels. These truly get rid of the order dependency, and they show the cost of the cold start for each implementation when compared to the warmed up runs.

Note the result for CPU M1 Pro, where the cold-start variant is faster than the warmed up one. That's just because the variance is fairly large in this case. I've attached a table where the experiment was repeated 10x, which shows that the cold start is in fact slower but that that slowdown is hidden by noise.

With his insight, the 2x factor between the two implementations is now gone. In hindsight, it was very suspicious that just the screen simulation should impact the entire GPSR run this much. I imagine that the runtime of GPSR is dominated by things like the backpropagation etc. much more than by the charge deposition in `Screen`. Hence, the differences between the two implementations are also small in the context of the entire runtime of GPSR.

reconstruction_6d_full_comparison

Screenshot 2026-06-18 at 15 17 01

jank324:
@roussel-ryan I'm not sure how you timed the two implementations against each other with GPSR, but is it possible that the 4.5x difference you saw was that warmup difference? That's the only place so far where I could reproduce a scenario where it looked like the tensor-based implementation was slower (even when it actually wasn't).

roussel-ryan:
@jank324 I ran the two methods against each other on the 4d reconstruction example and in separate jupyter kernels and got scattered results. At this point lets just use whatever implementation you think is best since I trust your benchmarking more than mine at the moment

jank324:
Here are the final benchmark results. If anyone wants to run the notebook themselves, it is added as part of this PR under `dev/benchmark/cloud_in_cell.ipynb`.

First on a MacBook Pro with an M1 Pro (10‑Core CPU, 16‑Core GPU und 16‑Core Neural Engine, 32 GB RAM):
Screenshot 2026-06-30 at 11 37 30

Second on a Maxwell cluster node with 2x AMD EPYC 7543 (each with 32 cores and 64 threads), 1007 GB RAM and 1x A100 with 80 GB:
Screenshot 2026-06-30 at 11 43 28

tests/test_cloud_in_cell.py
Hespe, line 1: Since we only test for 1D, 2D, and 3D inputs, the general fallback path for nD inputs is not tested at all

jank324: That's true. I just added a 4D test. Anything else?

Hespe: No, looks good otherwise

Hespe (approved):

PR #627, opened 2026-03-20
Only track non-scalar attributes in `UNVECTORIZED_NUM_ATTR_DIMS`

cr-xu:
## Description

- Remove default zeros in Beam.UNVECTORIZED_NUM_ATTR_DIMS and only track attributes with non-zero dimension in ParticleBeam and ParameterBeam; 
- Introduce a `defining_features` property in the same fashion as for Elements.
- Simplify some unit tests by using the `defining_features` property in beam classes to verify dtypes and devices are tracked properly. All the tested attributes are generateed from these definiing_features and the test was bit redundant.

## Motivation and Context

## Types of changes

## Checklist


cr-xu:
This one is actually ready for review since a while.

jank324:
Does this change really improve anything?

cr-xu:
It handles the `dispersion_x` y, px, py terms which are currently not covered, and probably will make sure that it never happens again in the future when we add more scalar beam quantities.
The current list of `UNVECTORIZED_NUM_ATTR_DIMS` is already super long!

jank324 (approved):

PR #636, opened 2026-05-11
Fix missing off diagonal terms in the `from_parameters` construction in beam classes

### LINKED ISSUE #634: `ParticleBeam.from_parameters()` takes incomplete cov terms
The `ParticleBeam.from_parameters()` construction method only takes partial covariance information and assumes zero correlation from cross plane, which will lead to unexpected errors.

In addition, `ParameterBeam.as_partricle_beam()` uses this construction instead of direction from the full cov matrix, so it will sometimes lead to error due to non SPD covariance

cr-xu:
## Description

Add off-diagonal elements in `from_parameters` construction method in beam classes; 
Use `from_distribution` construction when converting `ParameterBeam` to `ParticleBeam`.

## Motivation and Context

## Types of changes

## Checklist


Hespe:
This is also an issue with `transformed_to`, it neglects the off-diagonal terms in the covariance. Maybe we should fix that as well if we are at it.

tests/test_parameter_beam.py
```diff
@@ -22,6 +22,14 @@ def test_create_from_parameters():
...
         cov_ypy=torch.tensor(0.0),
         cov_taup=torch.tensor(0.0),
+        cov_xy=torch.tensor(1.0e-15),
+        cov_xpy=torch.tensor(-1.1e-15),
+        cov_xtau=torch.tensor(1.2e-15),
+        cov_pxy=torch.tensor(-1.3e-15),
+        cov_pxpy=torch.tensor(1.4e-15),
+        cov_pxtau=torch.tensor(-1.5e-15),
+        cov_ytau=torch.tensor(1.6e-15),
+        cov_pytau=torch.tensor(-1.7e-15),
```
jank324: This test is inconsistent in what it what can be set, what it sets and what it checks.

tests/test_parameter_beam.py
```diff
@@ -196,3 +212,47 @@ def test_conversion_to_and_from_particle_beam():
     assert torch.isclose(
         original_parameter_beam.total_charge, reconstructed_parameter_beam.total_charge
     )
+
+
+def test_conversion_to_particle_beam_preserves_full_covariance():
```
jank324: This test duplications `test_conversion_to_and_from_particle_beam`. That test should instead be adapted.

tests/test_particle_beam.py
```diff
@@ -30,6 +30,14 @@ def test_create_from_parameters():
...
         cov_yp=torch.tensor(0.0),
         cov_pyp=torch.tensor(0.0),
+        cov_xy=torch.tensor(1.0e-15),
+        cov_xpy=torch.tensor(-1.1e-15),
+        cov_xtau=torch.tensor(1.2e-15),
+        cov_pxy=torch.tensor(-1.3e-15),
+        cov_pxpy=torch.tensor(1.4e-15),
+        cov_pxtau=torch.tensor(-1.5e-15),
+        cov_ytau=torch.tensor(1.6e-15),
+        cov_pytau=torch.tensor(-1.7e-15),
```
jank324: Same note on inconsistency as for `ParameterBeam` equivalent test.

jank324 (changes requested):

tests/test_parameter_beam.py
cr-xu: Good point. I just realized the high rtol and atol in the `test_conversion_to_and_from_particle_beam` was due to this missing off-diagonal term, now they match properly and we don't really need to check for individual parameters anymore.

tests/test_parameter_beam.py
cr-xu: I moved checks to all np.isclose. But not sure if it is what you're getting at.

cr-xu: I see, I added all off-diagonal elements now and tested below.

tests/test_particle_beam.py
```diff
@@ -23,15 +23,22 @@ def test_create_from_parameters():
...
+        cov_pxtau=torch.tensor(-1.5e-15),
+        cov_pxp=torch.tensor(1e-15),
+        cov_ypy=torch.tensor(1e-15),
+        cov_ytau=torch.tensor(1.6e-15),
+        cov_yp=torch.tensor(1e-15),
+        cov_pytau=torch.tensor(-1.7e-15),
+        cov_pyp=torch.tensor(1e-15),
+        cov_taup=torch.tensor(1e-15),
         energy=torch.tensor(1e7),
-        total_charge=torch.tensor(1e-9),
```
jank324, line 34: Why was `total_charge` removed?

cr-xu: Good point! It shouldn't be! It's probably just lost in copy-paste / tab-completion.

jank324 (approved):

PR #639, opened 2026-06-05
Patch emittance `NaN` and `0.0` conditions

jank324:
## Description

Under some edge conditions it is possible that floating point arithmetic causes the emittance values of a beam to become 0.0 or a very small negative value, causing Twiss parameters computed from them to become `NaN` or `Inf`. These conditions occur often enough to disturb ML processes. Once such a condition occurs, the gradients under that condition is useless anyway, so clamping to a constant numerically useful value is fine as well ... and at least avoids `NaN`s and `Inf`s.

## Motivation and Context

`NaN`s and `Inf`s coming from randomised Cheetah simulations, especially for Twiss parameter values.

## Types of changes

## Checklist


jank324:
@Hespe, you can have a look at what Copilot says, but I think we can ignore all three suggestions for now.

Hespe (approved):

PR #646, opened 2026-06-30
Add optional metadata fields to elements

### LINKED ISSUE #645: Add optional metadata fields to accomodate control system PVs and other usage
@roussel-ryan suggested to include PV names to the elements which will make the integration of Cheetah models as digital twin / virtual accelerator more convenient.

Heres's my suggestion:
We simply add an optional dictionary style attribute `metadata` to the Element base class, containing seriaizable fields that we can export to a json file (str, float, array, ...).

We make clear that simulation shouldn't be depending on it, and it can contain any extra information people want to store along with the lattice file.
We don't enforce any format on how the fields structure should be, but give templates / instructions in documentaion page.
It could look like:
```python
q1.metadata = {
    "control_system": {
        "type": "EPICS",
        "pv_base": "A:Q1:",
        "setpoint": "SetCurrent",
        "readback": "MeasCurrent",
    }
}
```

cr-xu:
## Description

Add an optional, free-form `metadata` field for every element. This can be passed either during construction or at later stage.

It is intended for annotation and will not be used for simulation. Currently, the main use case is for integration with the control system and building digital twin applications.

The metadata field is not included in the `defining_features` so it does not invalidate the transfer-map cache.
`clone()` will deepcopy the metadata fields.

For now, we don't enforce any template, but provide a suggested structure on in the documentation `element_metadata.md`. In the long term, we will make this conform with the standards in [PALS](https://github.com/pals-project/pals).

## Motivation and Context

## Types of changes

## Checklist


cr-xu:
@roussel-ryan  @jank324  @Hespe  You can take a look, also check what can be improved in the documentation: https://cheetah-accelerator--646.org.readthedocs.build/en/646/examples/including_metadata.html

roussel-ryan:
LGTM

Hespe:
Looks good. My only question would be on metadata for virtual elements like `SpaceChargeKick`, but I guess we want to have that anyway for consistency reasons.

jank324:
> Looks good. My only question would be on metadata for virtual elements like `SpaceChargeKick`, but I guess we want to have that anyway for consistency reasons.

Yeah, I think it needs to be consistent. Control system info is only one way we can think of using `metadata` right now. It could be used to save all kinds of notes, or even for LLM agents to make annotations.

jank324:
I've gone over everything. I made a few smaller changes, but overall it looks good to me too, and I think this PR is ready for review.

One bigger thing I did was to add a schema for how the control system annotations might look with DOOCS and a mapping from Cheetah element properties in mind. I like that it includes a mapping that a digital twin can use to figure out which property of the Cheetah element maps to which DOOCS address (or EPICS PV). I don't know if this should also include annotations on things that might not map directly to a Cheetah property. An example on a DOOCS magnet would be the `PS_ON` property. I don't know if the digital twin should be aware of that. @roussel-ryan, do you have an opinion?

Otherwise, I would propose that we modify the EPICS examples somewhat in this image. Like keep the `pv_base` entry, but change the `readback` and `setpoint` entries to a dictionary that maps from Cheetah properties to PV names. But I don't know if we can actually do that, because for the quadrupole only a current PV is given, which does not have a corresponding property on the Cheetah element.

Maybe the mapping also shouldn't be from Cheetah property to control system property, but rather the other way around, from the control system property to the Cheetah property. In the latter case, one could set the value of the entry to `None` to say that no corresponding property is available in Cheetah.

I would aim to make these schema examples reasonably nice now as a best practice for people reading them, but also keep in mind that they are definitely a living and moving target and therefore likely to change in the future as we actually implement digital twins or adapt PALS.

Please let me know your opinions @cr-xu, @roussel-ryan, @Hespe.

cr-xu:
Regarding the property <-> channel mapping, many facilities operate in direct current / voltage units instead of the normalized strength (used in simualtion). 
I think making it a dictionary with controllabe properties as keys is a good idea, prevents the hard-coded rules for different elements
For my case specifically, the reality looks more like:
 ```python
 q1.metadata = {
   "control_system": {
     "pv_base": "A:Q1:PS:"
     "properties": {
       "k1": {
         "setpoint": "SetCurrent",
         "readback": "MeasCurrent",
         "calibration_file": ["AQ1.excitation"],  # A excitation file used for interpolating current -> integrated strength value
   }
 }
 ```
 where I also define the calibration files (or a simple linear factor).
 Then I'll have an additional util file that defines the bidirectional conversion functions:
 `current_to_k1` or `k1_to_current` that takes the calibration file, beam energy etc.
 
I guess the only complaint is this makes it a bit more verbose as most elements just have one knob to change, but would make sense for cavities (field, phase, ...) etc.

cr-xu:
And, a `"status": "PS:ON"` makes a lot of sense and should be included in the digital twin. Optional check that only updates the model when everything is connected and up, or running fully detached.

jank324:
@cr-xu @roussel-ryan, how do the schemas look now? If they are good, I will merge it.

Hespe (approved):

Hespe:
I am not sure if the suggested schemas are sufficient. We will probably see only when the first prototypes are in development. But since we do not impose any specific schema, this should not block the PR in my opinion.

roussel-ryan:
I think the example is good and flexible enough to handle future changes down the road. Let's not settle on a final structure until we play with it a bit / hear from the community

PR #658, opened 2026-07-15
Streamline option to silence `DirtyNameWarning` and fix to name sanitisation

### LINKED ISSUE #657: Raise `DirtyNameWarning` only if `sanitize_name` is not specified.
The various constructors of `Element` subclasses currently raise a `DirtyNameWarning` whenever the provided name is not a valid Python identifier and suggest to set `sanitize_name = True` to get rid of the warning.
However, even if you explicitly specify `sanitize_name = False`, which should be sufficient to show that the invalid names are as intended, you get the same warning.

I would suggest to change this behaviour to only raise a warning when `sanitize_name` is not specified.

jank324:
## Description

Automatically suppress `DirtyNameWarnings` when `sanitize_name` is explicitly set to `False`, and replace the sanitisation check with a Python builtin function in addition to other minor changes fixing a big where some non-sanitised names would not be recognised and/or sanitised correctly.

## Motivation and Context

Closes #657.

## Types of changes

## Checklist

### DIFF
--- a/cheetah/accelerator/element.py
+++ b/cheetah/accelerator/element.py
@@ -33,7 +33,7 @@ class Element(ABC, nn.Module):
     def __init__(
         self,
         name: str | None = None,
-        sanitize_name: bool = False,
+        sanitize_name: bool | None = None,
         metadata: dict | None = None,
         device: torch.device | None = None,
         dtype: torch.dtype | None = None,
@@ -44,12 +44,13 @@ class Element(ABC, nn.Module):
         if not self.is_name_sanitized():
             if sanitize_name:
                 self.sanitize_name()
-            else:
+            elif sanitize_name is None:
                 warnings.warn(
                     f"Dirty element name {self.name} is not a valid Python variable "
                     "name. You will not be able to use the `segment.element_name` "
                     "syntax to access this element. Set `sanitize_name=True` to change "
-                    "the name to a valid one, if you want to use this syntax.",
+                    "the name to a valid one, or `sanitize_name=False` to silence this "
+                    "warning.",
                     category=DirtyNameWarning,
                     stacklevel=2,
                 )
--- a/cheetah/accelerator/element.py
+++ b/cheetah/accelerator/element.py
@@ -355,12 +355,10 @@ class Element(ABC, nn.Module):
         alphanumeric_with_underscores = "".join(
             c if c.isalnum() or c == "_" else "_" for c in self.name
         )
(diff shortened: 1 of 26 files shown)


cheetah/accelerator/element.py
```diff
@@ -343,25 +345,22 @@ def split(self, resolution: torch.Tensor) -> list["Element"]:
...
     def sanitize_name(self) -> None:
         """
         Sanitise the element's name to be a valid Python variable name.
 
         Replaces characters that are not alphanumeric or underscores with underscores.
+        If the name starts with a number, an underscore is prepended.
         """
-        self.name = "".join(
+        alphanumeric_with_underscores = "".join(
             c if c.isalnum() or c == "_" else "_" for c in self.name
```
Hespe, line 356: Actually the `or c == "_"` can be removed, right?

```suggestion
            c if c.isalnum() else "_" for c in self.name
```

tests/test_elements.py
```diff
@@ -428,3 +429,23 @@ def test_consistency(element, beam_cls):
...
+
+    if sanitize_name is None:
+        # Check that the warning is raised
+        with pytest.warns(DirtyNameWarning):
+            _ = cheetah.Marker(name=dirty_name, sanitize_name=sanitize_name)
+    else:
+        # Check that no warning is raised
+        with warnings.catch_warnings():
+            warnings.simplefilter("error", DirtyNameWarning)
+            _ = cheetah.Marker(name=dirty_name, sanitize_name=sanitize_name)
```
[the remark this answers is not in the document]
jank324: Yeah, I don't want to bloat the test with that.

cheetah/accelerator/element.py
```diff
@@ -343,25 +345,22 @@ def split(self, resolution: torch.Tensor) -> list["Element"]:
...
-        ).strip("_")
+        )
+        stripped = alphanumeric_with_underscores.strip("_")
+
```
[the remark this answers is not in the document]
jank324: The idea wasn't bad, but decided to remove the stripping instead.

cheetah/accelerator/element.py
jank324: I think in theory you are right, but I feel like the current version of this code makes the intent clearer when you read it, right?

Hespe: Yes, agreed.

Hespe (approved):

PR #659, opened 2026-07-16
Add transverse focussing to the `Undulator` element

Hespe:
## Description

This PR adds a linear transfer map with transverse undulator focussing. So far, the `Undulator` did only model a drift section of the appropriate length.

## Types of changes

## Checklist


Hespe:
The consistency tests currently fail, but that is expected. I did not update the stored files yet, since I did not want to repeat that step in case there are other changes.

Hespe:
Also, Ocelot supports imaginary `Kx` and `Ky` (without real part only, though). Neither Elegant nor Bmad seem to handle that, atleast in their standard wigglers, so I decided to skip that for now. Just adding a couple of `.real` does not seem to do the job, even though that should properly switch to the hyperbolic `sinh` and `cosh` for imaginary `Kx`/`Ky`.

cheetah/accelerator/undulator.py
```diff
@@ -38,7 +42,9 @@ class Undulator(Element):
     def __init__(
         self,
         length: torch.Tensor,
-        is_active: bool = False,
+        period: torch.Tensor | None = None,
+        Kx: torch.Tensor | None = None,
+        Ky: torch.Tensor | None = None,
```
jank324: I think these should be named a bit differently. Capital `K` is very unpythonic. We could so something small, e.g. `kx` or maybe `horizontal_undulator_parameter` to be more descriptive and avoid using equation symbols as variable names? What do other codes do?

jank324 (changes requested):

cheetah/accelerator/undulator.py
Hespe: Ocelot has `Kx` and `Ky`, Elegant has `K` and Bmad `kx`.

Hespe: Its also a bit confusing, that `Kx` is responsible for vertical focussing and vice versa.

jank324: Let's go with `kx` then.

Is it the same in other codes, that `kx` is for vertical focusing and vice versa? Maybe this is because a horizontal undulator creates a vertical focusing effect? (I'm not sure a horizontal undulator creates a vertical focusing effect, I'm just guessing.

cheetah/converters/bmad.py
```diff
@@ -241,9 +241,13 @@ def convert_element(
...
             )
         elif bmad_parsed["element_type"] == "wiggler":
-            validate_understood_properties(shared_properties + ["l"], bmad_parsed)
+            validate_understood_properties(
+                shared_properties + ["l", "l_period", "kx"], bmad_parsed
+            )
             return cheetah.Undulator(
                 length=torch.tensor(bmad_parsed["l"], **factory_kwargs),
+                period=torch.tensor(bmad_parsed["l_period"], **factory_kwargs),
+                Kx=torch.tensor(bmad_parsed["kx"], **factory_kwargs),
```
Hespe: Not sure anymore if this is the right mapping between parameters.

cheetah/converters/elegant.py
```diff
@@ -350,6 +350,22 @@ def convert_element(
...
                 sanitize_name=sanitize_name,
             )
+        elif parsed["element_type"] == "wiggler":
+            validate_understood_properties(
+                shared_properties + ["l", "k", "poles"], parsed
+            )
+
+            # There are two poles in one period
+            length = parsed.get("l", 0.0)
+            period = 2 * length / parsed["poles"] if "poles" in parsed else 0.0
```
jank324:
```suggestion
            period = 2.0 * length / parsed["poles"] if "poles" in parsed else 0.0
```

cheetah/accelerator/undulator.py
```diff
@@ -54,22 +60,67 @@ def __init__(
...
-        tm[..., 2, 3] = self.length
-        tm[..., 4, 5] = self.length * igamma2
+        tm[..., 4, 5] = (
+            -self.length
+            * igamma2
+            * (beta.square().reciprocal() + 0.5 * (self.kx.square() + self.ky.square()))
+        )
+
+        spatial_frequency = torch.where(
+            self.period > 0,
```
jank324:
```suggestion
            self.period > 0.0,
```

cheetah/accelerator/undulator.py
```diff
@@ -54,22 +60,67 @@ def __init__(
...
+        self.register_buffer_or_parameter(
+            "kx", kx if kx is not None else torch.tensor(0.0, **factory_kwargs)
+        )
+        self.register_buffer_or_parameter(
+            "ky", ky if ky is not None else torch.tensor(0.0, **factory_kwargs)
+        )
+
+    @property
+    def is_active(self) -> bool:
+        return torch.logical_or(self.Kx != 0, self.Ky != 0).any().item()
```
jank324:
```suggestion
        return torch.logical_or(self.kx != 0.0, self.ky != 0.0).any().item()
```

cheetah/accelerator/undulator.py
```diff
@@ -54,22 +60,67 @@ def __init__(
...
+            * (beta.square().reciprocal() + 0.5 * (self.kx.square() + self.ky.square()))
+        )
+
+        spatial_frequency = torch.where(
+            self.period > 0.0,
+            math.sqrt(2) * torch.pi / (self.period * gamma * beta),
+            torch.tensor(0.0),
+        )
+
+        # Transverse focusing from vertical field (Kx > 0)
```
jank324:
```suggestion
        # Transverse focusing from vertical field (Kx > 0.0)
```

cheetah/accelerator/undulator.py
```diff
@@ -54,22 +60,67 @@ def __init__(
...
+        omega_x = spatial_frequency * self.kx
+        cos_omega_x = (omega_x * self.length).cos()
+
+        tm[..., 2, 2] = cos_omega_x
+        tm[..., 2, 3] = (omega_x * self.length / torch.pi).sinc() * self.length
+        tm[..., 3, 2] = -(omega_x * self.length).sin() * omega_x
+        tm[..., 3, 3] = cos_omega_x
+
+        # Transverse focusing from horizontal field (Ky > 0)
```
jank324:
```suggestion
        # Transverse focusing from horizontal field (Ky > 0.0)
```

jank324 (approved):

PR #667, opened 2026-07-22
Element merging

jank324:
## Description

As a counterpart to `Element.split` that splits elements into a sequence of shorter elements, we add a function `Element.merge` that merges consecutive elements into a single longer element.

## Motivation and Context

This helps deal with and improve compute speeds on some lattices, where large elements were split into sequences of smaller elements. We find this for example in the Elegant lattices of European XFEL. It also helps cleanup lattices, where a longer drift section is interleaved with markers. Removing the markers and then merging can speed up simulations quite significantly.

## Types of changes

## Checklist


cheetah/accelerator/sextupole.py
```diff
@@ -126,6 +130,28 @@ def is_skippable(self) -> bool:
...
+        return self.__class__(
+            length=self.length + other.length,
+            k2=self.k2,
+            misalignment=self.misalignment,
+            tilt=self.tilt,
+            tracking_method=self.tracking_method,
+            name=merge_element_names(self.name, other.name),
+            dtype=self.length.dtype,
+            device=self.length.device,
+        )
```
jank324, line 153: Actually, is merging elements with second order tracking valid?

cheetah/accelerator/quadrupole.py
```diff
@@ -273,6 +274,32 @@ def split(self, resolution: torch.Tensor) -> list[Element]:
...
 
+    def merge(self, other: "Quadrupole") -> "Quadrupole | None":
+        if (
+            self.tracking_method != other.tracking_method
+            or self.num_steps != other.num_steps
+        ):
+            return None
+
+        if not (
+            self.k1.equal(other.k1)
```
Hespe: This is unnecessarily strict. It would be ok to have different `k1` but those have to be weighted by the length.

cheetah/accelerator/quadrupole.py
```diff
@@ -273,6 +274,32 @@ def split(self, resolution: torch.Tensor) -> list[Element]:
...
+            and self.misalignment.equal(other.misalignment)
+            and self.tilt.equal(other.tilt)
```
Hespe, line 284: Would this not be prone to false rejections since we are comparing floats?

cheetah/accelerator/quadrupole.py
```diff
@@ -273,6 +274,32 @@ def split(self, resolution: torch.Tensor) -> list[Element]:
...
+        ):
+            return None
+
+        return self.__class__(
+            length=self.length + other.length,
+            k1=self.k1,
+            misalignment=self.misalignment,
+            tilt=self.tilt,
+            num_steps=self.num_steps,
```
Hespe: Why don't we add the two steps together? Than we do not need to enforce them to be the same? Of course, that would remove the computational effort benefit that we are hoping for, but that would anyways be mostly for linear tracking in my opinion.

cheetah/accelerator/sextupole.py
Hespe: It might not be, not sure. I guess better be safe than sorry and remove this again?

cheetah/accelerator/solenoid.py
```diff
@@ -136,6 +137,19 @@ def split(self, resolution: torch.Tensor) -> list[Element]:
             for _ in range(num_splits)
         ]
 
+    def merge(self, other: "Solenoid") -> "Solenoid | None":
+        if not (self.k.equal(other.k) and self.misalignment.equal(other.misalignment)):
+            return None
+
+        return self.__class__(
+            length=self.length + other.length,
+            k=self.k,
```
Hespe: Assuming a `Solenoid` is like a `Quadrupole`, we can probably do weight the `k` by length to allow for different values.

cheetah/accelerator/segment.py
```diff
@@ -311,6 +315,42 @@ def inactive_elements_as_drifts(
...
+                type(current) is type(next_element)
+                and current.name not in except_for
+                and next_element.name not in except_for
+            ):
+                merged = current.merge(next_element)
+                if merged is not None:
+                    current = merged
+                    continue
+
+            merged_elements.append(current)
```
Hespe, line 357: This only ever merged two elements that are consecutive, right? What about three or more instances of the same class? I think you just have to put this line into an else?

```suggestion
            else:
                merged_elements.append(current)
```

Hespe (changes requested):
Two general comments:

1. We should think about name sanitization in `merge`. I would argue for passing `sanitize_name=False` to the constructor within.
2. This does not work with nested `Segment`, right? Should we change this?

cheetah/accelerator/quadrupole.py
jank324: I guess. If we say if they are numerically close enough, the slight disturbance in the physics is fine with us, then comparing for closeness would be better.

docs/examples/custom_elements.ipynb
Hespe, line 1: I don't think that `sanitize_name=self.is_name_sanitized()` is a default. It disables sanitation exactly for those names that need to be sanitized.

I would rather go with `sanitize_name=False` or remove it. See also #669

cheetah/accelerator/segment.py
```diff
@@ -307,6 +311,49 @@ def inactive_elements_as_drifts(
...
+        if except_for is None:
+            except_for = []
+
+        merged_elements = []
+        current = self.elements[0]
+        for next_element in self.elements[1:]:
+            if not type(current) is Segment:
+                current = current.with_consecutive_elements_merged(
+                    except_for=except_for
+                )
```
Hespe: Just a thought: Should recursive merging be default but optional?

cheetah/accelerator/segment.py
```diff
@@ -307,6 +311,49 @@ def inactive_elements_as_drifts(
...
+            being mergeable.
+        :return: Segment with consecutive mergeable elements merged.
+        """
+        if except_for is None:
+            except_for = []
+
+        merged_elements = []
+        current = self.elements[0]
+        for next_element in self.elements[1:]:
+            if not type(current) is Segment:
```
Hespe: Wait, this gets executed exactly when `current` is *not* a segment?

Hespe: Also, I think this branch drops `next_element` without adding it to the resulting merged segment.

cheetah/accelerator/segment.py
```diff
@@ -307,6 +311,49 @@ def inactive_elements_as_drifts(
...
+                current = current.with_consecutive_elements_merged(
+                    except_for=except_for
+                )
+            elif (
```
Hespe: I think this `elif` should just be an `if` to ensure `next_element` is not silently dropped when `current` is a `Segment`.

cheetah/accelerator/segment.py
Hespe: And should a `Segment` merge its subelements if it is in the `except_for` list? I would probably skip merging in that case.

tests/test_merge.py
Hespe, line 1: We have added some more features to `merge`, like recursive merging in `Segment`, the `except_for` argument, and merging of three or more element. Should those be tested as well?

Hespe (changes requested):

docs/examples/custom_elements.ipynb
jank324: I changed that in both `split` and `merge`.

cheetah/accelerator/segment.py
jank324: > Just a thought: Should recursive merging be default but optional?

I think only if and once someone requests that as a feature.

jank324: > And should a `Segment` merge its subelements if it is in the `except_for` list? I would probably skip merging in that case.

Yep. Fixed it.

cheetah/accelerator/segment.py
jank324: Yeah ... I think now it makes more sense? Does it?

cheetah/accelerator/segment.py
jank324: Ah ... hmmm ... I (think I) solved it differently now. But is a bit tricky to think through.

Hespe: Hmm, but this will not merge two consecutive `Segment`, right? This will only merge the subelements now.

Hespe: I think the recursive merging should be separate from the top-level merging.

1. Check if `current` is a segment, then do a recursive call and merge subelements
2. Check if `current` and `next_element` can be merged, then merge, otherwise proceed.

Those two checks could be independent, right? Right now we have mixed them, which prevents merging consecutive `Segment`

jank324: I'm a little confused now. Isn't that (almost) exactly what the code does right now?

I just made a small fix to do the advance from `current` to `next_element` if the merge was attempted but not successful.

Maybe on a sidetone: Two consecutive sub-segments should not be merged according to our current specification. So merging would be applied to them internally, but as Segments they would remain separate fro each other.

tests/test_elements.py
```diff
@@ -461,3 +461,4 @@ def test_element_no_internal_dirty_name_warning(element):
 
         _ = element.clone()
         _ = element.split(torch.tensor(1.0))
+        _ = element.merge(element.clone())
```
jank324, line 464: Why `clone`?

jank324: Aaahhhh

jank324: Nevermind

Hespe (approved):
One might consider a test that merging and splitting preserve the metadata as intended, but that is not urgent I would say.

PR #678, opened 2026-07-28
Add a linear tracking method for `TransverseDeflectingCavity`

### LINKED ISSUE #677: Add a linaer tracking method for `TransverseDeflectingCavity`

cr-xu:
## Description

Added `linear` tracking_method for the `TransverseDeflectingCavity` and also made it the default method. Implementation follows the [ocleot conterpart](https://github.com/ocelot-collab/ocelot/blob/master/ocelot/cpbd/elements/tdcavity_atom.py).

## Motivation and Context

## Types of changes

## Checklist


CHANGELOG.md
Hespe, line 1: PR number should be #678

tests/test_transverse_deflecting_cavity.py
```diff
@@ -4,6 +4,47 @@
...
+        voltage=torch.tensor(2e6),
+        phase=torch.tensor(0.125),
+        frequency=torch.tensor(3e9),
+        tracking_method="linear",
+    )
+
+    outgoing = tdc.track(incoming)
+
+    assert outgoing.s == incoming.s + tdc.length
+
```
Hespe: I think instead of this new test, we should rather modify `test_parameter_beam_tracking_with_device_and_dtype` in `test_elements.py`. It currently excludes `TransverseDeflectingCavity`, we just have to move it to the other category that is excluded if tracking is not linear.

tests/test_transverse_deflecting_cavity.py
```diff
@@ -4,6 +4,47 @@
 import cheetah
 
 
+def test_transverse_deflecting_cavity_off():
```
Hespe: This should probably be parametrized on `tracking_method`?

tests/resources/consistency_expected_outgoing/TransverseDeflectingCavity_ParticleBeam_drift_kick_drift.pkl
Hespe, line 1: This file should be the same as before, right?

Hespe: Same for the two `*_inactive.pkl` files, actually.

Hespe (changes requested):
Looks good overall, only a few small remarks.

It would be great to have a comparison test to Ocelot like the other ones in _test_compare_ocelot.py_.

tests/test_transverse_deflecting_cavity.py
cr-xu: Yes, the suprising thing is with tdc off and `drift_kick_drift` it actually doesn't reproduce drift (also why the inactive file was changed in below)... I'll look more closer to it.

cr-xu: Actually, it's not suprising now that I rethink about it.
In TDC, when it used `drift_kick_drift`, it drifts l/2, applies kick, and drifts l/2 again. The Drift using `drift_kick_drift` applies drift once, so the correct replacement test would be two drifts with half the length ,and that reproduced perfectly.

tests/resources/consistency_expected_outgoing/TransverseDeflectingCavity_ParticleBeam_drift_kick_drift.pkl
cr-xu: > Same for the two `*_inactive.pkl` files, actually.

The inactive is not the same as before, per reason I replied above.

cr-xu: This file is newly generated, but yes, it's the same as the "active" one before.

tests/test_transverse_deflecting_cavity.py
cr-xu: Done

Hespe:
On a slight tangent: Why is the TDC `phase` given in terms of `rad / 2π`, not `rad`? And for `Cavity` its degrees. I think we should unify that at some point (not in this PR, though), @jank324

Hespe (approved):

cr-xu:
> On a slight tangent: Why is the TDC `phase` given in terms of `rad / 2π`, not `rad`? And for `Cavity` its degrees. I think we should unify that at some point (not in this PR, though), @jank324

I also stumbled on this... I think it's originally set in this way to be consistent with the bmad convention, but I agree it's kind of confusing.

PR #680, opened 2026-08-11
Vectorized histogramming

### LINKED ISSUE #351: Feature request: Support vectorisation for `Screen.reading` histogram
Screens are great outputs for training against, and doing them in batches would be preferred!

jp-ga:
Add vectorization for `Screen.reading` histogram (see #351 )

Allows fast computation of vectorised histograms when automatic differentiation is not required (model evaluation, for example).

## Types of changes

## Checklist


jank324:
I ran some benchmarks just to check if this makes sense given the presence of the new Cloud-in-Cell option and compared to the PyTorch-native implementation (in terms of speed).

## Compare PyTorch-native histogram to the implementation from this PR
Screenshot 2026-08-19 at 17 00 31

## Compare the different methods in Cheetah at the state of this PR
Screenshot 2026-08-19 at 17 01 09

cr-xu:
I'm confused.. the native pytorch histogram crashed for large batching or the custom vectorized version?

otherwise the speed comparison is pretty good. It's faster than cloud-in-cell for large screen, which justifies its usage?

jp-ga:
> I'm confused.. the native pytorch histogram crashed for large batching or the custom vectorized version?
> 
> otherwise the speed comparison is pretty good. It's faster than cloud-in-cell for large screen, which justifies its usage?

It seems that the pytorch histogramdd was the one crashing on GPU because the numbers of the custom vectorized version match with the 'Compare the different methods in Cheetah at the state of this PR' bottom plot. There might be an issue with the histogramdd runs on GPU because the ones that did run were significantly slower than running on CPU. Either way, the custom vectorized histogramming seems to perform better in most usable cases

jank324:
Yeah, so the PyTorch-native implementation crashed on MPS. This is an issue I've run into before. Very specifically the MPS implementation seems to require unusually large amounts of memory, so it's quite easy to make it crash on large inputs. But this isn't an issue with CUDA, so not a general GPU issue and probably not so relevant for most use cases.

I agree that generally it looks like the new vectorised implementation outperforms the PyTorch-native one. Tow things still confuse me:
 - I would have thought the PyTorch-native one is faster because it should be just a single C++ kernel. But I guess it could be a somewhat neglected implementation. It was only added to PyTorch very late (after Cheetah), the MPS implementation has severe issues, it doesn't support vectorisation, and overall it is not exactly something widely used in neural network training implementations.
 - **More confusing** is the "Non-vectorized 1M" case, which is an outlier in that the new implementation is suddenly and only here a lot slower than the original. I don't understand why that is, but it is reproducible.

I did some scans to kind try and understand that outlier. Interestingly, they look more mixed than I would have expected from the bar plots above.

Screenshot 2026-08-20 at 09 43 55

jp-ga:
The main reason why histogramdd is significantly slower in the vectorized case is because it isn't actually vectorized. It flattens the batch dimensions. That was the motivation to do the custom code.

The results you show are interesting though. I suggest we merge and then report these observations as an open issue for the future.

jank324:
> The main reason why histogramdd is significantly slower in the vectorized case is because it isn't actually vectorized. It flattens the batch dimensions. That was the motivation to do the custom code.
> 
> The results you show are interesting though. I suggest we merge and then report these observations as an open issue for the future.

Yes, of course I am aware of that. The case that confuses me is specifically a non-vectorised one.

jp-ga:
> > The main reason why histogramdd is significantly slower in the vectorized case is because it isn't actually vectorized. It flattens the batch dimensions. That was the motivation to do the custom code.
> > The results you show are interesting though. I suggest we merge and then report these observations as an open issue for the future.
> 
> Yes, of course I am aware of that. The case that confuses me is specifically a non-vectorised one.

I see. I'm not sure what could be causing that behavior for high number of particles. 

The tradeoff of the new implementation would be: 'native' vectorization which is significantly faster for large batch size (>10) but significantly slower for small batches when N>~50k. I still think it is worth it since the evaluation time is small when the batch size is small anyways, and we gain significant speedup for large batches (where the nominal evaluation time is large) + robustness in MPS GPU evaluations. Further optimization of the vectorized histogramming could be left as a separate future issue.

cheetah/utils/histogram.py
```diff
@@ -0,0 +1,99 @@
...
+    x2_edges = _bin_centers_to_edges(bins2)
+    bins_x1 = x1_edges.numel() - 1
+    bins_x2 = x2_edges.numel() - 1
+
+    ix1 = (
+        (torch.bucketize(x1_flat, x1_edges) - 1).clamp(0, bins_x1 - 1).long()
+    )  # (B, N)
+    ix2 = (
+        (torch.bucketize(x2_flat, x2_edges) - 1).clamp(0, bins_x2 - 1).long()
+    )  # (B, N)
```
cr-xu, line 86: Am I reading correctly that this clamp will put particles outside of the bin ranges to the outer edges? I.e. if you have a smaller screen than the beam, you will actually have a lot of intensity at the ourter rim

jp-ga: Good catch. I fixed it. see lines 70-81

